PR #42, opened 2015-01-10
Bug fix - cannot update ccv code for vault payment

rastographics:
Hard to believe no one else had this problem, so maybe I'm doing something wrong. But whenever I try to update my ccv code on a payment method that is saved, i was getting a blank value due to this code. 

Even if pf.CCV had a value, it would never get sent along to the StoreInVault routine.

You may have this by design, I just wanted to point out in case it's a bug.


drmohundro:
Hi @rastographics, thanks for this pull request! Because of some changes that I just pushed, though, your changes aren't necessary because we've removed the `MaskedCCV` completely. Per PCI, the CVV isn't ever allowed to be stored, so it is now only passed through to the payment processors for fraud purposes. Even payment processors aren't supposed to store it when using vaults, so when `StoreInVault` is called, it would only be used for ensuring it is a valid card.

I hope this makes sense. Feel free to comment if you have any questions!

rastographics:
Makes perfect sense, thanks.

PR #43, opened 2015-01-22
Bluepay Gateway added

rastographics:
This adds compatibility for BluePay payment processor (our church's payment processor). 
The admin settings that would be added are bluepay_accountId and bluepay_secretKey
TransactionGateway would be BluePay.

Added a dependency to the project CSVHelper, to help process the batch transaction data returned by BluePay's api. I hope this is OK.

Not sure how you like to handle dbschema changes -- I made a script that can be run on existing databases to add the Null field used to store BluePayVaultId.

Let me know if there's anything else needed, I would love to see this go live so our church can start using the online giving!


davidcarroll:
Thank you Joel,

We'll take a look at this next week and get it incorporated into production.

CmsData/Finance/BluePayGateway.cs
```diff
@@ -0,0 +1,359 @@
...
+                        "BluePay failed to save the credit card info for people id: {0}, message: {1}; transactionID: {2}".Fmt(
+                            person.PeopleId, response.Message, response.TransactionId));
+
+                //save for future
+                paymentInfo.BluePayCardVaultId = response.TransactionId;
+
+                paymentInfo.MaskedCard = Util.MaskCC(cardNumber);
+                paymentInfo.Expires = expires;
+            }
+            //TODO: Handle bank accounts too (not just credit cards)
```
drmohundro: Were you planning to add code for ACH vault support? I noticed `PayWithCheck` is implemented, but not this.

drmohundro:
Thanks @rastographics! Some questions:

- I commented on the related code, but were you planning to add code to support ACH (e-check) vaults? Credit card vault payments are there, but not ACH.
- Related to ACH, what about implementing returned checks? Does BluePay have a reporting API to get returned e-checks?
- Are there any test credentials to call BluePay with? The code is looking for `bluepay_accountId` and `bluepay_secretkey`, but our other gateways have hard-coded *test* credentials to make it easy for everyone to use.

rastographics:
* We do not use ACH for our church, so I did not implement yet. I noticed there is a property you have for gateways that allows ACH to be disabled so I set that property to disable ACH. But if you would like me to add the ACH functionality for complete-ness, I could do that.
* I don't know about the reporting for bad checks because we never deal with e-check, but again, if you need me to implement that I can find out.
* Bluepay issues a sandbox credentials on a per-user basis, so I don't have anything that could be hard-coded. But I simply set the IsGatewayTesting (I can't remember exact param name right now, but it's something like that) to true and the live bluepay credentials will only function in a test mode. (no sandbox credentials needed).

I hope I understood your comments correctly. Basically, do you want the ACH functionality included even though we won't be using it ourselves?

Thanks, looking forward to this.

drmohundro:
That makes sense... we'll go ahead and merge it as-is for now with the understanding that ACH processing won't work with BluePay for now. I'll put another comment on here once our changes are published to production.

Thanks!

drmohundro:
FYI, this is now in production. Thanks again @rastographics!

rastographics:
Thanks a lot.

Having a problem though, it seems the NoEChecksAllowed setting has no
effect on the production build of our live account (on my test machine,
setting it to true causes the Bank Account option to be hidden on payment
page).

Is there something I'm missing?...

davidcarroll:
Joel,
The NoEChecksAllowed setting had an extra space at the end of the name. I deleted the setting and recreated it and it works now.

PR #46, opened 2015-05-11
Speed up SendPeopleEmail for large List

davidcarroll:
Take a look at this to see if it makes sense or could be improved. I am using two DataContexts and the inner one is disposed after each recipient. The outer database context will contain any changes to the EmailQueueTo object.


drmohundro:
Looks good to me! :+1:

PR #47, opened 2015-07-10
Incorporate EmailReport from GSBCfamily Fork into BVCMS Master 

hkouns:
Originally coded and tested in Feb 2014 ... bvcms#27
I have integrated with the latest codebase  and retested.

EmailReport is designed to be very similar to EmailContent, except that
the body of the email is generated by a python script instead of being
pulled from an static file

See additional [documentation in this PDF](https://c4265878.ssl.cf2.rackcdn.com/gsbcfamily.1507101236.SendReport.pdf)


davidcarroll:
Heath,

This looks like a really nice idea and feature. Thank you. I will take a closer look this weekend. Here are a few things I'd like to see changed:

* We need to get you to use our documentation format instead of a PDF. So that it can be maintained in our bvcmsdocs repo
* We have gone away from the ceremonial code at the top of our Python scripts to a classless format. In other words, we are no longer requiring
```
  import System
  import System.Text
  from System import *
  from System.Text import *
  class mastertestscript(object):
      def Run(self):

```
* We have a much nicer way of capturing the output of a script instead of using StringBuilder, we just use the Python print statements and capture the standard output.

I'll send you some examples.

hkouns:
David,  
Thanks for looking at the feature.

Regarding the documentation, I had planned to write documentation for the feature and submit in the standard format.  I am currently working to get the tool chain installed.  I will respond to your other email with questions about that.

I just wanted to go ahead and get the pull request submitted... in case there are other changes before starting the formal documentation.  But I also wanted to have some form of description of the function and testing.

The print statement sounds much cleaner.  It doesn't sound like those changes will affect the pull request.  It will just make the script files nicer...right?   I look forward to the examples.

davidcarroll:
You will need to make a different call to the Python script that renders your report since there won't be a class nor a Run method. I'll just make those modifications after I merge the pull request so you can see what I did. We'll want to put some sample Python code in the documentation for sure.

Thanks for doing this. I understand the principle of pushing data to the minister rather than asking them to login. Some of the things we are working on with the app is Push Notifications and using "no login" links to take them directly to pages in the system. Those could be reports, Tasks, Contacts etc.

PR #48, opened 2015-07-14
Overloading QueryList with Sort Order

hkouns:
The quickest method I could find to perform a search was to insert a .ToList() after the Take(1000) followed by OrderedBy(parameter) like this:

       return db.PeopleQuery2(s).Take(1000).ToList().OrderBy(t=>t.LastName).ThenBy(t=>t.FirstName)

Is there a better way to do this?

For the overloaded QueryList function, I added a string for the orderby parameter and a bool for direction:

         QueryList(object s, string orderbyparam, bool ascending)

The orderby parameter will accept one of the following strings for now:
"age", "birthday", "name"

If the given parameter is not found in the switch statement then it defaults to an un-ordered list.

I thought about adding the orderby parameter and the Direction to the dictionary recently implemented for EmailReport .... but I was concerned that it would affect the API interface.

Let me know if this is sufficient for now.


davidcarroll:
You should use a pattern like:
https://github.com/bvcms/bvcms/blob/master/CmsWeb/Areas/Manage/Models/UsersModel.cs#L111-L129

The QueryList is giving you a IQueryable<Person> which can then be used as above.
You will want to apply the Take(1000) at very last.

hkouns:
David, 
Thanks for the pointer... updated to new structure.

hkouns:
David, 

I am a bit confused now... I thought I did update the code to using the linq2sql query enumeration pattern you referenced.  The only difference I see, is that you used a single IF statement and 2 Switch statements.  while I used a single switch statement and multiple IF statements.  Is that the readability improvement you are referring to?

or maybe I didn't add it to the pull request correctly... It says to  *"Add more commits by pushing to the master branch on GSBCfamily/bvcms."*    which is what I did... and the file changed on this pull request now shows the new code... (see below)

Also, one clarification, I am drawing a distinction between age and birthday... birthday would be used to make sure that all of the persons are sorted according to the month/day of their birthday. 
So that I can show everyone with birthdays tomorrow, then everyone with birthdays the next day, etc.


var q = db.PeopleQuery2(s);

            switch(orderbyparam.ToLower())
            {
                
                case "age":
                    if (ascending)
                        q = from u in q
                            orderby u.Age, u.LastName, u.FirstName
                            select u;
                    else
                        q = from u in q
                            orderby u.Age descending, u.LastName, u.FirstName
                            select u;
                    break;
                case "birthday":
                   if (ascending)
                       q = from u in q
                           orderby u.BirthMonth, u.BirthDay, u.LastName, u.FirstName
                           select u;
                   else
                       q = from u in q
                           orderby u.BirthMonth descending, u.BirthDay descending, u.LastName, u.FirstName
                           select u;
                    break;
                case "name":
                    if (ascending)
                        q = from u in q
						    orderby u.LastName, u.FirstName
						    select u;
                    else
                        q = from u in q
                            orderby u.LastName descending, u.FirstName descending
                            select u;
                    break;
                
            }
            return q.Take(1000);

davidcarroll:
Apparently I did not see your refactored code. Not sure but I see it now.  Looks good. And I see what you are doing with the age sort. Good idea. Sorry for the confusion. I'll merge this now.

PR #50, opened 2015-07-16
Fix for team scores not updating

drmohundro:
Opening this for @steven-bvcms


davidcarroll:
I like easy, one line fixes. Thanks Steven. As you can see we are doing this with a pull request and I think that is the way we should work going forward. You can still do the branch, but this way I don't have to download the branch.

steven-bvcms:
Sounds good.  I'll do pull requests.

PR #55, opened 2015-07-30
InsertDraft Replacement Code

hkouns:
I added an Email Replacement code in the form of {insertdraft:Name_of-Draft}.
It pulls the HTML content from the body of a named draft email.

**Use Case:**  I can setup a Template for Women's Ministry (sisters of grace - SOG) in which I include a couple replacement codes in uneditable portions of the template.
I can include one for Upcoming Events ->  {insertdraft:SOG_Events}
I can then include another for the Contact Us section -> {insertdraft:SOG_Contacts}

Currently, if they use the template then they must recreate the entire list of upcoming events each time... or they keep saving drafts of old emails and reuse the draft to keep the list of upcoming events updated.

With this feature, the upcoming events can be maintained separately in the draft... I don't have to update the template... and they can start from the "template" file each time.

**Status:** I have tested the use case above and it works well.  If the draft  doesn't exist, then it returns a string "Draft content could not be found".  

Currently there is no limitation on by Owner of Draft.  I can see a benefit of this...a secretary could be keeping the Draft file of upcoming events updated separately from the person sending an email.

If you think we need to limit by owner of draft, then I will expand ContentOfTypeSavedDraft() to include a second parameter for the sender ID and then include another "where" condition to limit the owner.  
If this is needed: What is the senderID in the context?  and How do I include Team Drafts as well as those owned by sender?

*Note:* If this is accepted then I will start working on the documentation page for the email replacement code.

*Note:* I requested this pull request from GSBCfamily/HK_Develop... trying to keep it separate from the pull request for Email Org Assignment Notifications that came from GSBCfamily/Master


hkouns:
Note this idea was originally submitted via Support Request as gsbcfamily [5654]

davidcarroll:
Heath, another neat idea. But, again, we are indeed slammed with work right now and we will just have to postpone testing this until we can get caught up. Thank you.

hkouns:
@davidcarroll  thanks for the info... will look forward to it being tested.  
If you have time to think about the draft ownership issue I mention in the pull request, let me know your thoughts or desired direction ... I can be working on a solution (IF ONE IS NEEDED)

hkouns:
@davidcarroll, 
By the way... I meant to mention that I tested this on a local copy of the GSBCfamily database that was backed up to our Rackspace Cloud Files location.  Thanks for getting that setup.  
It was nice having the full database to test on and not having to setup and contrive test conditions in a blank database.

davidcarroll:
Cool about the database download. We have to download databases to work with locally all the time. We can point to production, but only when trying to debug the issue. Not good to write to the database in production from the dev box.

davidcarroll:
Heath, I really like this idea but I am struggling with using the content of another email (draft) to insert into a main email template. The only thing that bothers me is that the draft is going to have the full `<html><head></head><body></body>` set of tags. so there will be an `<html>` embedded into another `<html>` tag. In actuality, browsers and email clients may be forgiving enough to ignore that bad markup. But there still could be problems, furthermore the idea of knowingly doing that kind of markup is bothersome.  

In the past, I have tried to introduce the concept of an HTML snippet that would not have the `<html>` tags but only `<div>` etc. But the implementation of that was very kludgy and not at all automatic or intuitive.  The HTML editors we use automatically insert the `<html>` tags. You can configure them to not do that, that is where the difficulty comes in, because you have to know that it is a snippet when the editor is loaded on page load, and not on save.

So, this pull-request may get more complicated before it makes it to production.

hkouns:
David... thanks for the feedback.

When testing the concept I had gone to "emails" and opened one that was previously sent and switched to HTML view.  All I saw was a `<DIV>` tag, so I assumed I was safe.

Also,  I just converted a couple of the test emails to HTML: (one with the inserted HTML and one with the replacement Codes)

http://gsbcfamily.org/wp/wp-content/uploads/2015/08/WithInsertedHTML.html
http://gsbcfamily.org/wp/wp-content/uploads/2015/08/WithReplacementCodes.html

When you look at the HTML it looks like it is just wrapped in a DIV... for instance the {addsaveddraft:SOG_Events} code is replaced with:

```HTML
<div style="max-width: 600px;" bvedit=""><table width="216" height="37"><tbody><tr><td valign="middle"><a href="http://localhost:51611/Display/EmailBody/166#"<strong>Upcoming Events</strong></a><br></td></tr></tbody></table><p><br></p><table border="0" cellspacing="1" cellpadding="1"><tbody><tr><td><h4><strong>Date</strong></h4></td><td><h4><strong>Time</strong></h4></td><td><h4><strong>Location</strong></h4></td></tr><tr><td><p>Friday, Jan. 16<br>  Monday, Jan. 26<br>  Saturday, Feb. 7<br>  Monday, Feb. 9<br>  Monday, March 9<br>  Saturday, March 21&nbsp;<br>  Monday, April 13<br>  Monday, April 27<br>  Monday, May 11th</p></td><td>6:30 pm<br>  6:30 pm<br>  5:00 pm<br>  6:30 pm<br>  6:30 pm<br>  8:00 am<br>6:30 pm<br>  6:30 pm<br>  6:30 pm</td><td><p>Kick-Off, GSBC Chapel<br>  1st Meeting, GSBC Fellowship Hall<br>55+ Valentine Dinner, GSBC<br>2nd Meeting, GSBC Fellowship Hall<br>3rd Meeting, GSBC Fellowship Hall<br>Spring Gala: Accessory &amp; Craft Sale<br>4th Meeting, GSBC Fellowship Hall<br>5th Meeting, GSBC Fellowship Hall <br>Finale, GSBC</p></td></tr></tbody></table></div> 
```

However, I do see that following the block of code is this block... I don't know where is comes from:
```
<!-- Visual Studio Browser Link -->
<script id="__browserLink_initializationData" type="application/json">
    {"appName":"Internet Explorer","requestId":"d42f0e8977c743179ab9e5a3a4bff112"}
</script>
<script src="http://localhost:50460/80e203b5d81f4bf3856bb2715718d79e/browserLink" type="text/javascript" async="async"></script>
<!-- End Browser Link -->
```

Do you think the Tags you are referencing were already stripped out?
could we search the returned string and replace the `<html>` tags of concern?
Do you think the "Visual Studio Browser Link" would be there in the production version?

davidcarroll:
The new Froala editor makes all the HTML appear on a single line which does not display well on a previously sent email. To get around this, we "pretty print" the HTML code for display, we are using HTML Tidy (an old open source project on http://tidy.sourceforge.net/). That actually will clean out the extra tags. But I don't trust it to do this except for display purposes, not wanting it to mess with the HTML in this way before sending the email out. So the email that get's sent in your scenario will have the extra `<html>` tags. You can only prove this by looking at the string right after you have done the email replacements. It is hard to do a "view source" on email content after it arrives at the client too. An email is not like a web page displayed in a browser, so to speak.

What I am thinking about doing is using the HtmlAgility pack or a Regex to get the innerHtml of the `<body>` tag if it exists.

hkouns:
oh ok ... Now I understand where HTML Tidy is being used... i had seen a couple commits about that a while back.

The emails that I attached were actually converted from eml to HTML using an online converter... dont know what that does to the code.

I was wondering if Regex could be used in this application... In fact, before I got fooled by the HTML Tidy rendering of the email,  I had considered searching for for the Tags and replacing them with a `DIV` ... but then when I looked I "thought" there was nothing to replace.

davidcarroll:
Download Thunderbird email client and you can open the eml files directly.

hkouns:
I do have Thunderbird ... I just converted this time thinking it might give a better representation of the html.

davidcarroll:
I discovered that our Edit Drafts function actually produces HTML without the full page <HTML> tags. So that was a moot point.  

I did move the InsertDrafts function to do it's work before the code replacements occur on every recipient. This way 1) the content will not have to be loaded from the database every time and 2) if there are any replacement codes in the draft itself, they will work as normal.

hkouns:
This is great!  Thanks!  And I like the change you made... it makes it even more powerful.

PR #56, opened 2015-07-31
Fix Org Member Sort

hkouns:
**Bug Fix**: Originally submitted via TPS Support Request  (gsbcfamily [5717])

**Solution**
The case condition in the Switch statement in ApplySort() should be "Date of Birth"
instead of "DOB" to match the column heading in
CmsWeb/Areas/Manage/Views/OrgMembers/List.cshtml
In the current live build, since it does not find "Date of Birth" it goes to the default which sorts by Name.
``` 
   default:
   case "Name":
```
**Testing**
I compared performance using same conditions on the life database and a downloaded database with the new build.

### DIFF
--- a/CmsWeb/Areas/Manage/Models/OrgMembersModel.cs
+++ b/CmsWeb/Areas/Manage/Models/OrgMembersModel.cs
@@ -209,7 +209,7 @@ namespace CmsWeb.Models
                             orderby om.Person.Name2
                             select om;
                         break;
-                    case "DOB":
+                    case "Date of Birth":
                         q = from om in q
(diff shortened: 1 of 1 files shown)


davidcarroll:
Thanks Heath for fixing this!

davidcarroll:
Heath, by the way, we just started using tags in Git to indicate what commit and when we published. You can see those in the Branch dropdown > Tags tab. 

The build date on the About page is kind of flaky and not always reliable.

hkouns:
Thanks for the info David…

I had seen Mo's commit on tags, but I have not tried to figure out where to find the tag information.

PR #58, opened 2015-08-13
Daily delete special tags

davidcarroll:
Was doing this on every new Session (every two hours), but if people were composing an email for over two hours, the session would expire and the Special email tag of their recipients would be gone. Wanting to make sure we are doing this right so let's discuss.


drmohundro:
Couldn't this still affect users who are composing an email at 11:45 PM but don't try to send until 12:05 AM?

Maybe it could be a rolling window instead? Something like:

```cs
if ((dbsessionstart?.Created ?? DateTime.Today) <= DateTime.Today.AddHours(-rollingWindowSize)
{
    // recreate here...
}
```

davidcarroll:
Maybe we should only delete email tags that are more than 24 hours old. That would be the rolling window and then we can forget about the midnight thing. Ostensibly, they will have either abandoned or sent their email by then. Then it could go back to just being when a Session Starts.

drmohundro:
I think that's fine - that follows the general expectations of session expiration (if there is any) more closely anyway.

PR #74, opened 2015-11-18
Python report on blue toolbar

hkouns:
This feature allows Python Reports to be included on the Blue Toolbar similar to the method used for SQL Reports by adding a line in the CustomReports text file in Special Content:

```
<Report name="DataVerification" type="PyScript" />
```

An ActionResult was added to the Reports Controller for PyScript.  That function populates a special tag called *"LatestPeopleFromCustRepContext"* with the people in the current context from which the blue toolbar is used (e.g. People Record, Org Members, Query Result, Tag, etc.)  The Tag is Cleared each time prior to population. 

The ActionResult also checks to see if a Query Function called "QueryCustReportTag" is included in the Python Script.
If it is NOT included, the Python script is not executed and an error message is provided.
If it is included, the Python script is executed and the HTML output is presented to the PyScript View.

When the Python Script is executed, it calls the "QueryCustReportTag" function which returns an IEnumerable<Person> list of the people in the *"LatestPeopleFromCustRepContext"* Tag.  The IEnumerable is limited to 1000 records.

For testing: 
- I verified it returned the proper people in multiple contexts (People Record, Org Members, Query Result, Tag) using my [DataVerification](https://gist.github.com/hkouns/08ace1b7d03f47f59649) Script.  I simply replaced the call to QueryList2 with a call to QueryCustReportTag.

- I update to the latest code at BVCMS/master and then tested again.

- I also used the Handlebars template script that David provided and verified that it worked.  (I like the way handlebars works....)

_NOTE: This Pull Request ONLY included file changes related to this feature.  However, you will see extra commits from when Nathan issued a pull request to GSBCfamily/BVCMS/master for a feature we are working on to add CC email capability.  Unfortunately, I merged it... but then I went back and reverted the commit on master and moved that work to a branch.  All of that happened before I made a new branch for testing this feature._


davidcarroll:
Heath, thank you for doing this. I think this is a very nice feature to compliment what we are doing with CustomReports. 

You will see in the push I am about to make that I made a few mods to avoid using a Special Temporary tag. Having the Guid for the query itself obviated the need for a Tag. So it is now passing the Guid through the dictionary object.  Also, I renamed the QueryCustReportTag function you had to BlueToolbarReport since the implementation was no longer using a tag.

hkouns:
Excellent!   Thanks so much.  Passing the Guid was exactly what I _WANTED_ to do...I knew that was the best approach but I couldn't figure out how to do it in the time I looked at it.    I am still not really familiar with the Dictionary Implementation.  (Need to improve my knowledge there.)

davidcarroll:
Yes, deleting the branch is the right thing. It does not delete any of the commits or code that was in the branch.

PR #75, opened 2015-11-20
Adding CC box on email

hkouns:
@nehresma and I have been working on a feature to add a CC box to the email form.  Nathan has done most of the work.

The primary use case is so that we can CC an address of someone or a email_list not in BVCMS.  For instance, if we send an email on behalf of the finance team and want each person to be able to reply to the email list for the finance team. 

At first, we wanted to add them to the CC field...but this would cause the CC'd person to receive too many emails.

Our final solution was to send a copy of the email to each address on the CC line (comma separated) These email do not get email replacement codes addressed.   These emails also do not get tracked in BVCMS since they are not associated with a peopleID.  Then we add each address to the Reply-To Field.  

I also added a note under the CC box on the form to describe its function and I added a line of text to the body of each email for "Full Disclosure" to the recipient and to inform them that the addresses are in the Reply-to field.  Currently it is at top and bottom for my testing.

Currently there is one bug that needs to be fixed....as long as the email addresses in the CC field are valid, then everything works correctly... however, if the email is invalid (e.g. "invalidemailtext") then the cc list is reset to null and the email is treated like normal...so noone on the cc list gets included.

What are your thoughts about this feature?  should we make any changes?

We eventually plan to add a button to the To: field and allow you to add extra recipients as needed...these will be added via the normal Search Add Dialog and thus will be valid PeopleIDs so the emails can be tracked.


hkouns:
[Here are a couple Screenshots](https://c4265878.ssl.cf2.rackcdn.com/gsbcfamily.1511201237.AddedCCscreenshots.pdf)

hkouns:
Also… For clarification, the modifications to the "To:" field will be handled in a separate branch and a separate pull request

davidcarroll:
* I understand why you might want to include a few extra people on the list, but I am not sure I understand the use case for sending to someone outside the database. 
* I would not want to put the Full Disclosure text on the page for every email since this will be a rarely used feature so I would hide all of the new UI elements in a single div and cause it to be shown when you check a box to use that feature.

hkouns:
@davidcarroll , thanks for the comments.  Let me address them in order.

**Use Cases:**
Several come to mind.  Each of these examples, I could have used multiple times in the last 6 months.  In these cases, the primary reason is not just to add the CC'd person as a recipient necessarily but to add them as a Reply-To address or to provide a level of accountability by copying a Pastor or staff-member.  I think that the future feature of adding people to the "To:" field will be better suited for just adding recipients. (and they would be in the database)

   *Use Case 1:*  A ministry leader in charge of a mission trip is sending out an email and wants to CC one of the Pastors. They primarily want the Pastor copied so that the recipients are aware that the pastor got the same email.  They also want the recipients to be able to reply to both the pastor and the ministry leader at the same time.  In this case, the pastor would be in the database...but I don't really want to limit the scope to addresses in the database.
	
   *Use Case 2:*  The church treasurer sends out the Church Budget to the active members for review.  She wants to CC the Finance Team email list address so that people can reply-all with questions. In this case I would prefer not to have the list-serve email address in the database as a fake person.
	
   *Use Case 3:*  The youth pastor is coordinating an event with a local venue and wants to CC the manager of the venue so that everyone is on the same page and so that parents can reply to the manager with information. In this case the manager is definitely NOT in the database
	
   *Use Case 4:*  This may be the most common example.  There are co-coordinators for a certain ministry or event (e.g. Children's Camp)  and when one of the coordinators send out an informational email they want both coordinators to be copied and get the responses.
	
**User Interface:**
I think there was a misunderstanding related to "Full Disclosure Text" and I will address it below. In regards to the UI, I agree that hiding the CC box and associated instructions would be beneficial... if for no other reason than to keep the UI clean and brief for use on mobiles.  We will start working on adding this.  I could see a similar benefit for hiding the "Schedule" box which I personally would use a lot less than the CC box.

**Full Disclosure Text:**
I was referring to the text that I added to the top of the email body that said: "This email was CC\'d to the email addresses below and they are included in the Reply-To Field."  The reason that I included it was that usually when someone is CC'd on an email the primarily recipient is aware that it was sent to someone else.  Since we couldn't put their name in the CC field, I thought that by adding a note to the body of the email, at least the recipient would know who else got it. *Do you think this is necessary?*

I have gone back and forth between just calling it "Reply-To" instead of "CC" or ""CC/Reply-To".  Do you have a thought on this?

davidcarroll:
Thank you. Let me get the rest of the team here to weigh in on this to see if there are any other ideas.

hkouns:
@davidcarroll ,
I appreciate the consideration and welcome additional thoughts and input. 

I added two more commits to the pull request and I posted a comment with representative images on each commit.

- The first one addresses the UI and provides for collapsing the DIVs around the CC box as well as the Schedule.  (treated separately)
- The second commit fixes the issue with an individual improperly formed email address in the CC box causing the entire CC line to be treated as empty.

Please let me know if the team has any other questions/input/concerns. @nehresma and I will try to answer any that come up.

Thanks again...

AndrewTriplett:
Heath,

I really think this is a very useful feature. We do receive a lot of questions from users wanting to reply to an email that originated from our system.

A common scenario that we're presented with is a Sunday school class that is planning an upcoming event. They need to send out an email, but it will be an ongoing conversation. Currently this must be done outside of our system.

In this scenario, with your feature in place, it may be nice to have a "copy to CC" field option. For example, if a leader needs to email 10 individuals for ongoing planning, they would have to copy the emails in the To field to the CC field. Giving them a one click option would streamline that.

Thoughts?

davidcarroll:
Andrew, let's discuss internally first. The thing you are asking about is unrelated to Heath's pull request. The reply thing is not the list serve that we have talked about. Think about when we send an email to 15,000 people, That would be impossible do do a reply to that many the way Heath has it implemented. The way that sort of thing is with a list server where you email to a single email and everybody get's notified.

hkouns:
Andrew,

Thanks for your thoughts on this.  I can certainly see the need for what you are describing.

However, I think it would need to be implemented in a different way.  I started by calling this a “CC” field but it is really more of a “Reply-To” field.   I don’t believe that you would want to add a “Large” quantity of addresses to the Reply-To Field.
Also, unless someone is using an email client that’s supports showing all headers like the Reply-To field (such as Thunderbird) they will likely not know who was “copied” until they hit Reply.     That is why (in the pull request)  I added the “footer” note to the bottom of the email message body that lists who is in the Reply-to field.

If you were willing to limit the scope of the “Copy to CC” feature, we could easily add the function but then only show the option if there are less than ~10 people in the “To:” field.   (and then you would have to decide where to put the threshold… 8 or 10 or 12 or???)

I just saw David’s email popup… I agree that a better method (longer term and a lot more work) would be along the listserve lines.  Maybe you could have the ability to associate a unique Email Address with an org.  Then you could treat each org as a miniature email listserve.   The unique address could certainly be added to the Reply-to field.  In fact, when you email org members you would probably want the ability to toggle the feature on or off.  I could see a checkbox for  “Allow Recipients to Reply to all Org members?”

Heath

davidcarroll:
Heath, 
We have all (including Karen) had a chance to review this idea and understand it and like it. We are going to move forward with it. I will pull the request to a separate branch and publish it for testing. Because of the Thanksgiving week, it may be two weeks before that is all done and ready to publish to production. Thanks for doing this. Blessings to you and yours. And tell Nathan @nehresma we appreciate his work on this too.

hkouns:
Thanks.  Glad you like it.

The timeline is fine.  Have a great Thanksgiving!

Heath

karenworrell:
Heath & Nathan,
I like the new feature! I've been testing and a found a couple of things I need to tell you about:

1) The Send button should be disabled as soon as you click to send the email. That's the way it currently works without the new cc/reply to code. We don't want people clicking it again and accidentally sending their email more than once.
2) When a recipient replies to the email, it goes only the those in the cc/reply to field, not to the original sender.

Have a blessed Thanksgiving!
Karen

hkouns:
Thanks Karen... 

Regarding:
1)  hmmm... We will have to look at that.   We didn't intentionally modified anything about the Send button so it looks a little troubleshooting is in order.

2) oops... Nathan had code in that did that.... I removed it when testing problem domains and forgot to add it back.

I also need to check the visibility of the "Schedule email" Checkbox and make sure it is only visible by the appropriate users.

karenworrell:
Thanks Heath. I have not finished testing, so I'll be sure to look at schedule email.

hkouns:
Karen,
I looked at the code a bit more tonight. I think I have fixed the original sender not getting the reply (Issue 1).  Also, I verified that the Schedule Email checkbox is not shown if you do not have the proper role.  (Issue I mentioned  in last response -- I had already fixed it and forgot)

Regarding the send button ... could you describe in more detail what you are seeing?  When I test,  the send button is immediately "dimmed" and you cannot click on it... and then it is hidden by the modal results window.   I tried to get a screen capture of what I am seeing:
send disabled

karenworrell:
Heath,
Once I send the email, a dialog pops up letting me it has been sent. When I click 'ok' to close the dialog, the Send button is still active. I can continue to send the email over and over.

hkouns:
Karen,  
Perhaps I am still not understanding...

What you describe above seems to be consistent with how it currently works for me on tpsdb.com.  When I send an email on the production system (gsbcfamily.tpsdb.com) the modal "results" dialog box blocks access to the Send button... but I can see that it turns from grey to dark blue in the background.  As soon as I click 'ok' on the dialog, I can click Send again.   
productiontest

I also pulled the CC Box branch code back down and tested it locally and I get the same behavior.

LocalTest

karenworrell:
Heath
You are correct! I'm sorry. I thought we had disabled the 'send' after you send the email.

davidcarroll:
Karen, it used to be disabled. Must have lost that when we released the new UI last May.

karenworrell:
David, I'll add a Trello card re disabling the Send button after sending, if we want to put that back in.

karenworrell:
Heath,
The 'reply' is working as it should. The email goes to the cc'd addresses as well as the Sender. Thank you! 

However, I've been trying to figure out how you are tracking the sending of these types of emails in the database, and I'm not finding anything. Even though the cc'd addresses are displayed for the recipients, I really think we also need to capture this info in the database, whether or not the email addresses are not connected to a people record. 

We track Parent 1 and Parent 2 when we use the CC Parents feature. Perhaps you can come up with a good way to display the CC'd addresses using your new feature.

Thanks for all your work on this!

hkouns:
Karen,
Originally, the thought was that it would not be tracked since the CC'd email addresses may not be in the database.  However, I certainly think that it would be wise to have a historical record of the addresses that were CC'd.

To that end I have pushed another Commit (https://github.com/GSBCfamily/bvcms/commit/f54fa2418c43e67269260146b043b000a1f58881) to facilitate this tracking.
- It requires a column (CClist) be added to the EmailQueue  Table.  
- When an email is sent, the a string of the CC Email addresses is stored in the CClist field.
- The string only contains the Validated email addresses ... improperly formed addresses are not stored
- When the Email Detail page is shown, a note is displayed under the "From" address stating that the email was CC'd to the following addresses...
- The note is not shown if there were no CC addresses

Tracking CC list

Do you think this is an acceptable way to track the CC'd addresses?

karenworrell:
One more question, Heath. You mentioned a new 'column'. Is that in a table in the database itself? The addresses themselves do not display in a column.

hkouns:
Yes…I am storing the string in a new column in the EmailQueue table in the database.

I put it there because it is being treated similar  to the “FromAddr” … or the “FromName“
Rather than trying to fit the addresses into the Recipient list since they are not tied to people records.

hkouns:
Superseded by PR https://github.com/bvcms/bvcms/pull/86

PR #76, opened 2015-11-27
Add FmtZip to Python Events

hkouns:
Needed to properly format 9 digit zip codes (add dash before last 4
digits)


hkouns:
Trying to improve formatting in the data verification Report.

**Test Results:**

*Address without FmtZip function call*
without

*Address with new FmtZip function call*
with

davidcarroll:
I also added FmtZip to the Handlebars helpers since these are the preferred way to do HTML templates.

This has been published to production.

hkouns:
Thanks David.  I look forward to learning more about Handlebars.  I am in the process of converting a couple reports now.

PR #78, opened 2015-12-07
Sql to Python Report

hkouns:
I was looking for a way to generate a list of Giving Units from the Church Database.  In a related email correspondence @davidcarroll  correctly stated that the best way to do this was using a SQL Report so that we have "_all of the combining data logic in SQL so that it is performant on large databases_"  Here is the [Example GIST](https://gist.github.com/davidcarroll/ee176fb60174429fa831)

This produced the correct fields and data that I was looking for, however it was lacking the formatting I desired for the report.  (my goal is for our Admin Assistant to be able to run the report and print it directly.)

I started looking for a way to import the results of the SQL script directly into a Python Report.  If there is already a way to do this in the codebase, I was unable to find it... or maybe I didn't understand what I found. ;-)

I found that there is a way to directly covert an IDataReader object to an IEnumerable or list.  This idea is what is covered by this pull request.  I found that the column headings are maintained and that I can access all the data in the Python report. It appears to return a list of DataRow objects.  This allows me to generate an HTML report that is formatted like I desire and the office can print without updating an offline excel file: (this report is used by the counting team on Sunday.)
table

The only issue that I found, was a difference between using Handlebars and the old _print '...'_ method.
When I used the old _print '...'_ method the key names are accessible directly  (_print q["PeopleId"]_) and the report runs very quickly.  Here is the related [_print '...'_GIST](https://gist.github.com/hkouns/5b3c7d320837524c2d19)     It was run against a slightly modified [GivingUnits](https://gist.github.com/hkouns/2f672cedfc3da0676db8) that pulls in PeopleID's and puts a lower limit on giving.

However, when I tried to covert it to a Handlebars template, I could NOT find a way to access the values directly. I tried things like _{{Name2}}_, _{{"Name2"}}_, _{{this/Name2}}_, etc. 
I found that I first had to iterate through the results of _QuerySqltoList(...)_ and map the values into a list of simple class objects.  See clip below:  (here is the full [Handlebars GIST](https://gist.github.com/hkouns/e988675d78968b3dd8a8) )

```python
sqlpeople = q.QuerySqltoList('GivingUnits',"")
people=[]

for q in sqlpeople:
    x = SimpleClass()
    x.Name2 =  q["Name2"]
    x.PeopleId =  q["PeopleId"]
    x.Spouse =  q["Spouse"]
    x.SpouseId =  q["SpouseId"]
    x.ContributionOptionsId =  q["ContributionOptionsId"]
    people.append(x)
   
body = model.RenderTemplate(template, people)
```

Then I passed the new list to RenderTemplate and I was able to access the fields directly in the _{{#each this}}_ section by using the key name: _{{Name2}}_

Unfortunately this seems to increase the processing time by roughly an order of magnitude over the  _print '...'_ approach.  (e.g. ~20sec vs <2sec on a report with ~150 rows)   I have not yet tested with other response sizes.

Any guidance on how to access the values directly in handlebars... or a better way of doing it all together would be appreciated.

What additional testing would you like to see?


davidcarroll:
No need to do the loop. The {#each} does that. The following should work
``sqlpeople = q.QuerySql(GivingUnits)``
then
``body = model.RenderTemplate(template, sqlpeople)``

hkouns:
I did that first … and it went through the loop the proper number of times  (I printed out the index to verify)
However,  I was not able to access the data …  I tried things like {{Name2}}, {{“Name2”}}, {{this/Name2}}  etc.      How should I format the parameter request?    I think it is complicated by the fact that it is a DataRow object.

Going through this extra loop was the only way I could get the data into a form I could call it in the {{#each}} section.  I can access the data by using _q["PeopleId"]_ BEFORE rendering the template...

hkouns:
And to clarify, the [Handlebars ](https://gist.github.com/hkouns/e988675d78968b3dd8a8) version of the report still uses the {{#each}} section.  It is just that it seems like I had to put the data into a simpler form for processing.    I would much prefer to eliminate the extra loop and just call the parameters directly in the {{#each}} section   ... i just couldn't make the values appear.

davidcarroll:
I'll rewrite it for you to show you how it is done. Give me a few days

hkouns:
ok...thanks

davidcarroll:
I just ran this one against your database, works fine. I did have to change SpouseId to SpouseID as the names are case sensitive.

https://gist.github.com/davidcarroll/8dbb4578a69e081cd2de

hkouns:
Awesome!!  Works Great!  Thanks.
I had played with "QuerySql" some  but I was't setting up the GivingUnits variable correctly.  I guess this is the part about "... I didn't understand what I found..."  

I will cancel the Pull Request.  Thanks again.

Issue #79, opened 2015-12-11
Help on Handlebars Helper


hkouns:
@davidcarroll ,
I am working on a couple Handlebars helpers and I have run into an issue and need some guidance.

The first I am working on is to add a Helper to duplicate the InOrg Python Event.

When I set it up like the "IfEqual" helper then it works properly:

Code
```
            /// IfInOrg in the form of:  {{IfInOrg pid orgId}}
            Handlebars.RegisterHelper("IfInOrg", (writer, options, context, args) =>
            {
                var pid = args[0];
                var orgId = args[1];

                var om = (from mm in db.OrganizationMembers
                          where mm.PeopleId == pid.ToInt()
                          where mm.OrganizationId == orgId.ToInt()
                          select mm).SingleOrDefault();

                if (om != null)
                    options.Template(writer, (object)context);
                else
                    options.Inverse(writer, (object)context);
            });
```
I can then use the following template:

```
{{IfInOrg PeopleID 48}}
   True
{{Else}}
   False
{{/IfInOrg}}
```

However, if I try to get the helper to return the string itself then I get an error:
"**Reached end of template before block expression was closed**"

Here is the code for that version:

```
         /// InOrg in the form of:  {{InOrg pid orgId}}
            Handlebars.RegisterHelper("InOrg", (writer, context, args) =>
            {
                var pid = args[0];
                var orgId = args[1];

                var om = (from mm in db.OrganizationMembers
                          where mm.PeopleId == pid.ToInt()
                          where mm.OrganizationId == orgId.ToInt()
                          select mm).SingleOrDefault();

                if (om != null)
                    writer.Write("Yes");
                else
                    writer.Write("No");
            });
```

How should I modify this helper so that it properly returns a string?  I have a couple other helpers that I would like to add... but they all follow a similar pattern.  The helper returns a string based on one of the arguments that is passed in.

_Note: In this particular case, I will probably use the "IfInOrg" version, because it gives the greatest flexibility in the template.  (based on the output, I can say True or Yes or checkmark)_

hkouns:
**Update:**  It appears to be due to the IF statement in the Helper... If I remove it and just add a writer.Write statement for testing then it works properly.
```
            /// InOrg in the form of:  {{InOrg pid orgId}}
            Handlebars.RegisterHelper("InOrg", (writer, context, args) =>
            {
                var pid = args[0];
                var orgId = args[1];

                var om = (from mm in db.OrganizationMembers
[...]

PR #81, opened 2015-12-16
Handlebars helpers

hkouns:
Carried over functionality of several existing Python Event routines into new Handlebars helpers  (e.g. InOrg, SundayForDate, FmtPhone, etc)
Added Optional Format feature to existing helper **FmtDate**
Also created several new Helpers  (e.g. "Now", "DateOffset", FmtAddress, etc)
I added error catching where ever I could.

I created a Script for Testing found [here](https://gist.github.com/hkouns/a1040d8d55391607da5e)
and the results can be seen [here](https://c4265878.ssl.cf2.rackcdn.com/gsbcfamily.1512161104.HandlebarsHelpers.png)
image

_Here is a full listing:_

**IfInOrg** in the form of:
  {{IfInOrg pid orgId}} {{/IfInOrg}}

**IfInSubGroup** in the form of:
  {{IfInSubGroup pid orgId "subgroup"}} {{/IfInSubGroup}}
subgroup is case sensitive

**Math** helper in form of:
  {{Math lval oper rval "fmt?"}}
ex. {{Math Total "/" 12}}
fmt is optional. Uses standard/custom dotnet format strings
ex. {{Math Total "/" 12 "C"}}

**FmtNum** helper in form of:
  {{FmtNum value "fmt"}}
ex. {{FmtNum Total "C"}}
fmt is required. Uses standard/custom dotnet format strings

**FmtAddress** in the form of:
  {{FmtAddress PID "type" lines}}
type: "F" for family or "P" for Personal or "Primary" for primary
Lines: 1, 2 or 3  - format address with <BR> codes to fit on 1, 2 or 3 lines

**Now** in the form of:
  {{Now "fmt?"}}
Standard output format is M/d/YYYY
fmt is optional. Uses standard/custom dotnet DateTime format strings

**FmtDate** in the form of:
  {{FmtDate Date "fmt?"}}
fmt is optional. Uses standard/custom dotnet DateTime format strings
default format is M/d/yyy

**WeekNumber** in the form of:
  {{WeekNumber Date}}

**SundayForDate** in the form of:
  {{SundayForDate Date "fmt?"}}
fmt is optional. Uses standard/custom dotnet DateTime format strings
default format is M/d/yyy

**SundayForWeek** in the form of:
  {{SundayForWeek year week "fmt?"}}
fmt is optional. Uses standard/custom dotnet DateTime format strings
default format is M/d/yyy

**MostRecentAttendedSunday** in the form of:
  {{MostRecentAttendedSunday progid "fmt?"}}
fmt is optional. Uses standard/custom dotnet DateTime format strings
default format is M/d/yyy

**DateOffset** in the form of:
  {{DateOffset date offset "type" "fmt?"}}
fmt is optional. Uses standard/custom dotnet DateTime format strings
default format is M/d/yyy

**FmtPhone** helper in form of:
  {{FmtPhone phone# "prefix"}}


hkouns:
_Note:_ I am also also working on a Helper in the form **{IfCond lval "symbol" rval}}   {{/IfCond}}**
where symbol could be one of: <,  >,  <=,  >=, &&, || 
but it will be a later Pull Request.

davidcarroll:
You have really gotten into this! thanks Heath. Let me know when you are ready for us to pull.

hkouns:
@davidcarroll , 
I went ahead and added:

**IfCond** in the form of:  
{{#IfCond lval oper rval}} {{/IfCond}}


**Compare** in the form of:  
{{Compare lval oper rval}}

The difference is that **IfCond** returns the *options.Template* or *options.Inverse* while **Compare** returns text "True" or "False"

I created a new Script for Testing [here](https://gist.github.com/hkouns/2586ff5b7b3e9bac63dd)
and the results can be seen [here](https://c4265878.ssl.cf2.rackcdn.com/gsbcfamily.1512162322.HandlebarsHelpers_IfCond.png)
image

Please go ahead and pull whenever you are able.

Thanks

hkouns:
Also, just a note (this should not prevent pulling the code as is)


Ideally you could nest the helpers like the example I show below ... but as I mentioned in https://github.com/bvcms/bvcms/issues/79 the sub-expressions feature does not seem to be working as I expected.  (Based on my understanding of Handlebars documentation... which could easily be wrong!)
```
{{#IfCond (Compare Total ">" 100) "&&" (Compare MaritalStatusId "==" 20)}}
   do something
{{else}}
  do something else
{{/IfCond}}

```

davidcarroll:
I believe you should be able to nest conditions. Please make a simple example that proves it does not work. Basically, the rendering is recursively called. Your example above would not work of course, but just have:
```
{{#IfCond Compare Total ">" 100}}
   {{#IfCond Compare MaritalStatusId "==" 20}}
     do something
   {{/IfCond}}
{{/IfCond}}
```
You should be able to accomplish what you need without the {{else}} part, just multiple ifs

hkouns:
@davidcarroll, I agree that you can *nest* if statements similar to the way that you show.   It works well for AND logic... but not so well for OR logic.   _(note: that in your example... "Compare" is not an argument to "#IfCond" ....  It should be {{#IfCond Total ">" 100}} )_

However, I guess I should have said that sub-expressions are not working.  I am talking about is what is described in this statement on the http://handlebarsjs.com/expressions.html website:
subexpressions

Based on this, I think that I should be able to replace lval or rval in the _{{#IfCond lval oper rval}}{{/IfCond}}_ call with a subexpression.

Similarly, I should be able to replace lval or rval in the _{{Compare lval oper rval}}_ call with a subexpression.

A very simple example would be to compare Monthly giving to a threshold.

```
{{Compare (Math Total "/" 12) ">" rval}}
```
In this example, the Math sub-expression should be processed first and the result fed into the Compare helper as the lval parameter (args[0])

As you suggested I created a quick test script that verifies both that Sub-expressions are not working and that nesting IfCond statements does work.
subexpressiontes

Actually, I just noticed that using two IfCond statements to represent OR did not really do what I was wanting...because it printed the text twice.

davidcarroll:
I think you should modify your SQL statements to deliver the data in a way that will simplify and obviate the need for complex branching requirements in your handlebars template. Let the SQL do the heavy lifting and let the handle bars simply display the results in a simple, non-complex way. 

You should also refer to the Handlebars.net project and if you want, create issues for him about what is not working like the handlebars.js project.

hkouns:
Agreed on both accounts... that is why I was suggesting it was a minor issue and shouldn't interfere with the pull request.

Actually I didn't know for sure which implementation you were using.  When I search Github for Handlebars.net and look at the primary project, I see that it is already listed as a bug.  However, he believes that the solution will be a major "breaking change" https://github.com/rexm/Handlebars.Net/issues/48 

Thanks

davidcarroll:
Heath, I pulled these so I could examine them and do some testing. I appreciate you doing all this work. 

However, I don't want to make handlebars helpers into a programming language in and of itself. I think much of what you are doing ought to be in building the data structure, either in Python or SQL or C# and not done in the handlebars helpers. I am concerned about multiple calls to the database for each row and how these helpers encourage this. Also, the coercion of values to double to do math and comparisons just does not feel right. 

I like adding the formatting capabilities, however, and I think I like the IfCond helper. I am going to use a few of these things, but I may decide to remove some for the sake of not encouraging bad programming practices on those who come behind to write templates. I am going to do more to illustrate best practices. 

My way of adding features thus far has been to work on an as needed basis and just in time approach. I do want to make the Python scripting as simple as possible and have features in the model that makes things easier to do. I just want to move into this slowly. Whatever we do during these early stages, sets the patterns that get cut in stone for the future, especially since these "programs" that will be using these features are not part of the project and thus become a "contract" that we are bound by to maintain compatibility with for the future. 

Also, since whatever bad practices we enable here could eventually sap much computational time and power on our shared servers. Thus far we have managed to keep all 330 of our churches on two servers, one for the web servers and the other dedicated to the database.

hkouns:
@davidcarroll , 
Thanks for the consideration. I understand your concerns.   I like the changes you made to combine the **fmt** helpers into one.   I knew that a couple helpers like **FmtAddress** and **IfInOrg** would be a stretch because of the the extra calls...but at the time it seemed easier than figuring out the SQL.

I am still trying to convert my data verification report over to handlebars.
I have now figured out a SQL script that gets all fields that I need and creates a column on the fly with a value based on the person's membership in a particular org.  [Full GIST is HERE] (https://gist.github.com/hkouns/369e6f737eaf586d2736)

Here is the code I am using to determine if they are in an org:
```
    CAST(CASE WHEN EXISTS(SELECT 1 FROM dbo.OrganizationMembers o WHERE  o.PeopleId = p.PeopleId AND o.OrganizationId = 145 )
    THEN 1
    ELSE 0 END AS BIT) AS InGSBCNews,
```
**Question 1) Is this the best approach?**

I still think the **#IfCond** would be useful because it could reduce the following snippet from the gist template to a single line:
from:
```                
                    {{#IfEqual DecisionTypeId 10}}Yes{{/IfEqual}}
                    {{#IfEqual DecisionTypeId 20}}Yes{{/IfEqual}}
                    {{#IfEqual DecisionTypeId 30}}Yes{{/IfEqual}}
                    {{#IfEqual DecisionTypeId 40}}Yes{{/IfEqual}}
                    {{#IfEqual DecisionTypeId 50}}Yes{{/IfEqual}}
```
to:
```
                    {{#IfCond DecisionTypeId ">=" 10}}Yes{{/IfCond}}
```

**Question 2) What do you think needs to be modified with the _ifcond_ helper to make it better?**




Some of the others (e.g. **{{Now}}**  ) would primarily be used in the header of a report.  For Instance, I have created a report that shows current Giving Units based on those giving in the last 6 months.  I want to include information at the top that says:
"This report represents donors for the 6 months prior to {{Now}}"

**Question 3) Can {{Now}} be incorporated?**
*Or is there a way to generate the data as a variable in the Python script and then pass it along with the SQL query results into the template?*



Finally,  since I am now using custom SQL within the Python Report the *BlueToolbarReport()* is not working and now calling the python script from the blue toolbar is failing *(note: I can add an unused call to the function just to trick it into running the rest of the scriptfor testing)*
...  I am looking at modifying **public ActionResult PyScript(Guid id, string report)...** in the ReportsController to look for both BlueToolbarReport as well as something like "@qtagid" in the text of the report...


**Question 4) What do I add to the SQL Script to make this work?**
For example should I add:
```
JOIN dbo.TagPerson tp ON tp.PeopleId = p.PeopleId
WHERE tp.Id = @qtagid
```

**5) What is the best method to fix the code it so that a Python Script called from the Blue Toolbar will only be processed on the correct people?**

davidcarroll:
Here is a rewrite of your report with my changes 
https://gist.github.com/davidcarroll/5c2d0f317a28ec7d6764

1. The org membership code looks fine

2. The IfCond is unnecessary since you can do it more succinctly in SQL
    ``CASE WHEN DecisionTypeId >= 10 THEN 'Yes' ELSE '' END``

3. There is already DateTime element on model that returns Now
    And yes, you can pass any data into the model with ``Data.x = 2``, then x is available
    You will need to pass in your query results too in this case.
    Data is passed in by default if nothing else is passed in.
    See how I added two things onto Data: Now and sqlpeople.

4. See the gist where I used @BlueToolbarTagId (that is something new I added today)

5. See the previous answer

Also, I got rid of some your IfCond 
and replaced them with joins to the lookup tables and a few CASE statements where you weren't using anything from the lookup tables.

I tried to clean up the HTML code a bit too. There were numerous misformed tags, missing closing tags, outdated attributes, etc. You should consider writing your HMTL with an IDE that checks these things for you. I would put a lot more work into making the report look better before I published it for other churches to use.

I hope this helps you.

PR #85, opened 2016-02-11
New Check-In Release Candidate 2

steven-bvcms:


davidcarroll:
Steven, you good with me publishing this to production? I assume this will not affect anything other than your new iPad Checkin.

steven-bvcms:
I've double checked it and it only affects the new Check-In, so you are good to publish to production.

PR #88, opened 2016-02-15
Pull CCs from db rather than from EmailController closure #87

### LINKED ISSUE #87: [New Email Features]  CC Feature not working on Scheduled emails.
Tested sending an email to myself as original Recipient... Added recipients Penny and @nehresma ...  then CC'd Logan (at) kounsfamily (dot) net.

Sent essentially the same email two ways:
1)  Sent Immediately... the email came through perfectly and all features worked.  (Additional Recipients get email, CC'd person gets email, CC'd address is added to ReplyTo,  "This Email was CC'd  email was CC'd..." note was added to body) 

2)  Scheduled for 5 minutes in the future.
 ^^^  The Schedule portion worked... it came through at the right time.
 ^^^  The Email went to the correct  Recipients...
 ^^^  The CC'd address was validated correctly... it is persisted and saved correctly with the Email Details.  (when you go to the  Detail page, it says the email was CC'd to Logan)

 xxx  HOWEVER, the CC'd Address does not actually get added to the Reply to Field.
 xxx  The "This email was CC'd..." note does not get added to body of email.
 xxx   A copy of the email is not sent to the CC'd address

Reference #86

nehresma:
Fixes #87.  Since scheduled emails run in a different process than the web app we should be pulling the CC list from the database rather than a closure.

### DIFF
--- a/CmsData/Email/Emailer.cs
+++ b/CmsData/Email/Emailer.cs
@@ -465,7 +465,7 @@ namespace CmsData
             return aa;
         }
 
-        public void SendPeopleEmail(int queueid, List<MailAddress> cc = null)
+        public void SendPeopleEmail(int queueid)
         {
             var emailqueue = EmailQueues.Single(ee => ee.Id == queueid);
             var sysFromEmail = Util.SysFromEmail;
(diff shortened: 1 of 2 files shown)


davidcarroll:
Excellent @nehresma! Thank you for fixing this so quickly!

hkouns:
I have just tested and it fixed all the issues I identified in #87. 
I also verified that it is still correctly validating emails in the Cc string and removing non-emails.

davidcarroll:
See my comments on #87 Published, and plan for removing developer only restriction.

karenworrell:
The CC'd message is displaying on emails even when that box is not checked and no one is being copied.

nehresma:
Ack!  That's a side effect of passing around an empty List now instead of null.  Fixing.  New PR coming momentarily.

PR #89, opened 2016-02-15
Only display CC message when CCing

nehresma:
Fix for the issue @karenworrell noticed in PR #88.  Since that PR is closed, I'm creating a new one with this fix.  I'm on Linux at the moment and haven't tested this, but it should work.

### DIFF
--- a/UtilityExtensions/Emailer.cs
+++ b/UtilityExtensions/Emailer.cs
@@ -45,7 +45,6 @@ namespace UtilityExtensions
             }
             if (cc != null)
             {
-               
                 foreach (var a in cc)
                 {
                     msg.ReplyToList.Add(a);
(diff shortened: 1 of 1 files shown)


davidcarroll:
I am publishing this to production websites now, but I will wait until later to update the batch service as that is more involved. So, we won't be able to test this for scheduled emails until after the batch service is updated, but it should work fine now for the website application. Thanks again for the quick turnaround @nehresma.

nehresma:
Sure.  Sorry for missing it earlier and causing a hassle for an extra deployment.

davidcarroll:
No problem, this is how it goes. The deployment is the easy part :-) at
least for the website. The batch service update is a bit more involved but
not hard. It is rare that I need to update the batch service.

PR #93, opened 2016-02-26
Transnational Api Fix

brianlankford1:
So turns out the documentation from TransNational was incorrect and didn't specify that "payment" parameter was needed, but it is required.  Went ahead and add it put into the credit card and ach refund and void requests just incase.


davidcarroll:
Thank you Brian!

PR #103, opened 2016-04-06
Allow dash in settings names

drmohundro:
The [documentation specifies settings](http://docs.touchpointsoftware.com/Administration/CustomHeader.html) such as `UX-HeaderLogo` but the prior regex wouldn't allow for dashes. This should fix that.


drmohundro:
It looks like the regex was added at db9e2ed6af3e6864bfe9fba6d012cdbb2ba94e3a.

PR #105, opened 2016-04-08
Login shell fix

brianlankford1:
Renamed custom layout to shell layout instead.
Also needed to update the controller where bootstrap needs to be injected in "shell" html and also fixed the TryLoadAlternateShell() code to be called even if you get a validation error because the ViewBag data has to be set again after post.


drmohundro:
@davidcarroll I don't think this PR is ready to go to `master` yet - we'll keep working on it.

davidcarroll:
Karen will need to thoroughly test this before we publish anyway. It needs
to go to the work site

PR #109, opened 2016-04-19
Mobile App attendence: Allow roster filtering by subgroup

rastographics:
Well our church really needs this feature...I probably should have asked about it first, but I've been wanting to dive into the source of touchpoint for a while to get a feel for it anyways. Any comments or suggestions are welcome of course. Also I can write some documentation if you desire to publicize this feature eventually.

**Use Case: **
Our sunday school classes each have 1 Teacher. Classes are divided into sub-groups for attendance purposes. Each sub-group has a group leader. The group leader is responsible for taking attendance for their own group using the Mobile App.

**Problem**
Some classes have a size of 30-50 people. The sub-groups are broken down to size of 10-15 people.
Each group leader has the entire class roster showing on the app when taking attendance. This makes it difficult for them to keep track of who they are responsible for marking present.

**Solution**
Using the built-in sub-groups feature of Organizations, the attendance roster in the app can be filtered to only show members who are in the same sub-group as the current user (person taking attendance).

**How to use it**
1. A new Organization setting is available under the Attendance tab called "Filter Attendance Roster by Subgroup". This setting must be checked for each organization that desires this feature.
2. Next, the user must be a member of 1 and ONLY 1 subgroup. If the user is not a member of a subgroup, or is a member of more than 1 subgroup, this feature will be ignored. (Maybe in future I could add way to intelligently handle multiple subgroups?...)
3. That is all. Now the Mobile App attendance roster will only show members who are in the same sub-group as the attendance taker.

**NOTES:**
1. Organizations may have people who are NOT in any subgroup. (I call them "groupless" people). Guests will also be "groupless" also. This feature handles groupless people by showing them only on the roster of the Org's Leader. They can be assigned to a subgroup any time and then they will show up on the appropriate user's roster.
2. I am hoping to also utilize this feature eventually (with a couple tweaks) to enable the group leaders to take attendance in our Main Worship Service, so that just their 10-15 people are visible on their app, but the Worship Service Organization would contain all 400 or so people that also are in the Sunday School classes.


rastographics:
I forgot I need to provide a public test server and video according to the pull request guidelines. I'll try to get that up tomorrow but it may be Thursday.

davidcarroll:
@rastographics I am glad you did this. I think this could be useful beyond your church.

**Suggestion:** Instead of limiting everybody in the org to a single sub-group, use pairs of sub-groups with the convention a leader having not only the sub-group she leads but also another sub-group with a prefix of "groupleader-" followed by the sub-group name. When creating the roll-sheet, you would look to see if the user had that prefix and then find everybody in the associated group. This way, not only would you not be limited by the one sub-group rule, but you could also have more than one leader per group who could take attendance.

I am sure that @karenworrell will be interested in testing this when you get it on a server. When it passes muster, she will write the documentation if you provide an outline for her.

rastographics:
Good suggestion. That change would also make it easier to do a check on the fly for any members part of a group that has no leader assigned, and lump them in with the "groupless" category. (to keep anyone from falling through the cracks). What do you think of "leader:groupname" syntax (to stay consistent with other group names mentioned in the docs that use colon).

I made my server public and it works fine over a browser, but the app can't connect to it. Is it because I'm using a non-80 port number?

davidcarroll:
There is some "phone home" code in the app that prevents it from working on
non Touchpoint clients servers. We can enable your server. I'll have to
check to see what that setting is and let you know.

davidcarroll:
I just need the URL for your test site and we can enable the App for that. Send it to david at bvcms.com

The leader:groupname scheme would be perfect

rastographics:
Sent you the test server information.
Works with the "leader:" naming convention.
If a user is leader of multiple subgroups, everyone from those subgroups will show up on rolllist.

Question: since this feature now requires the user to be in a subgroup starting with "leader:" before it will kick in, is it redundant to have the Org setting AttendanceBySubGroups? I feel like it would be better if I removed that now.

One less thing to setup and cause confusion, if it "just works" when they use the naming conventions that will be stated in the docs. Only downside is someone could trigger this functionality accidentally. 

But they would have to do these things before that happened: 
1. Org has subgroup with the "leader:" prefix 
2. Org has additional subgroup that has same name but without the prefix.
3. A person taking attendance is part of the "leader:" subgroup.

So to keep things less complicated for the user, we can depend on the naming conventions for now and if it ever becomes a problem for anyone, put the org setting back in.

davidcarroll:
The leader: prefix is likely to be not rare. So either we leave the setting or we change the prefix to be something less likely to be used normally. Maybe AttendLeader: or something like that. Also, there needs to be a way for someone to be able to see the attendance for the entire class and not have it filtered. Probably should do a quick video for Karen so she can see what you intended before she tests. Also, I like the idea of having a setting to prevent unnecessary lookups on people and OrgMemMemTags every time. In other words, only do the extra work if it is indicated with a simple flag.

rastographics:
OK Good point about performance and avoiding those extra calls. I agree need to keep the org setting.

I've still not been comfortable with the way I implemented this. Using the naming convention is quick and effective, but there are 3 issues I still have with it.
1. Makes 2 calls to OrgMemMemTags (plus joins MemberTag on each call).
2. List of Subgroups gets "muddied up" quickly with the extra "leader:" double per group.
3. Assigning leaders to groups is cumbersome for the user. (4 actions: switch to manage tab, create the "leader:" group, switch back, and assign).

By utilizing an IsLeader boolean column on OrgMemMemTags table, we can:
1. Make only 1 call to OrgMemMemTags (with NO join on MemberTag needed)
2. No "extra" subgroups. 
3. Assigning leaders simplifies to 1 action: ("Make Selected a Leader of Target Group")

The Action button on Sub-Group Members page would get 2 more actions;
* Make Selected a Leader of Target Group
* Remove Selected as Leader of Target Group

Also, the table of members on Sub-Group Members page would need extra column: "Group Leader for", and for each member would be a list of the groups they are leader of.

I know this will be more work to implement for me, but it seems less code smell, and especially less DB calls on a function like this that gets called a lot (and at simultaneous times from many users ie, Sunday AM), and better UX for managing.

Let me know if you like it and I'll crank it out.

> Also, there needs to be a way for someone to be able to see the attendance for the entire class and not have it filtered

Can you explain this a little more? Are you referring to seeing attendance on the app? Do you want the user to be able to choose on the fly from the app if the class gets filtered or not?

Thanks for the feedback

rastographics:
It was easy to do this and made code much simpler in the API controller. 
I would only have to make changes now for the Sub-Group management page.

Just wanted you to show you how much the api call gets simplified this way.

davidcarroll:
You have complicated the testing now by adding new UI components and database schema changes. I was fine with the sub-groups convention and your objections to that did not bother me at all.\

rastographics:
Is there any way I can take the testing work off of you in order to make it easier to pull in these changes? 

I made short videos showing the difference between the 2 implementations for the end user.
It really makes it so much faster using the second method.

[Using Naming Convention](https://vimeo.com/163613070/ca1b45dc12)
[Using LeaderId bool](https://vimeo.com/163613071/ff6c0a2e83)

For our own use case, this would save considerable time given the number of orgs and subgroups we would setup and be managing.

The second method is live on the server I sent you before. 
The first method is live on the same server, but port 4402.

If you gave me a list of test cases I could run through them, or also write the sql migrate script for the new column.
The 2 new UI buttons are exact same as the existing ones but point at different controller actions.

If this is not compelling enough, let me know and I'll roll back the pull request. Thanks for all your feedback.

davidcarroll:
We don't trust developers to test. Not even me. Not that developers don't test--they are just not good at testing all the crazy scenarios to which a user will subject the system. Karen has a knack for finding unforeseen bugs. Even then, our users find bugs. Developers are just not good testers as a rule.

My thoughts were more geared to not making this such an invasive project. Even though this will be a nice feature, I don't see it as mainstream. That is why I was happy with making it more of a hidden feature that will satisfy the churches that feel they need it.  (Remember we have 280 churches using the system and have been doing this for seven years and have not needed this yet.)

Regarding the sub-groups being cumbersome to manage, if you try to do it with the single member dialog box, for any significant number of people, I would agree. But have you tried the capabilities of the Update Members Dialog (multiple at a single time?). That works with the filter, or the check boxes and will also all you to create new groups at the same time.

Also, under the blue toolbar gear, there is a group member management page, which allows you to select, create, and assign people directly there using the Action menu. And you can do shift-select, click one checkbox then shift-select another away from that one and all between getting checked.

That is why I am just not feeling your pain on this.

davidcarroll:
Hey Joel, If you really want to make a big impact on the system, from the group management page, create a way to add multiple sub-groups from a dialog box textarea element. This would have a big impact on all users and will also help you address this.

rastographics:
Admittedly I'm not familiar with all the work that happens on the other side of development, the testing and migration, etc. so from my perspective I didn't see the changes as that invasive, just as a better implementation. 

But I understand that I _don't_ understand that side, so I'll just roll it back to the original implementation. :)

> Hey Joel, If you really want to make a big impact on the system, from the group management page, create a way to add multiple sub-groups from a dialog box textarea element.

My goal isn't to have a big impact, but I like trying to solve problems, and enjoy working on this project since it is something we are using more and more as a church. Can you describe a little more of what you are looking for regarding this idea? And how it would address the original idea of this PR?

davidcarroll:
I thought you were complaining about all the extra steps needed to create
the leader sub-groups. I was just suggesting a one step for all of them at
once.

Regarding the impact, I am speaking of you adding a specific and likely to
be misunderstood UI elements to sub-groups which are used extensively for
lots of other purposes. So I was in favor of keeping the hidden,
configuration and convention approach more in line keeping an advanced use
case. Keeps support requests down for clients that begin to assume behavior
that is not intended.

About the need for a leader to see all attendees, did you realize that the
desktop meeting page also uses the filtered rollsheet, not just the App?
This means no leader with a sub-group will ever be able to see the entire
roll.

rastographics:
> About the need for a leader to see all attendees, did you realize that the
desktop meeting page also uses the filtered rollsheet, not just the App?
This means no leader with a sub-group will ever be able to see the entire
roll.

I did not realize this. But I see the `@FromMobile` parameter on dbo.RollList is only TRUE when called from MobileAPI, so I can ignore the subgroup filtering if `@FromMobile` is false.

As to your point:

> So I was in favor of keeping the hidden,
configuration and convention approach more in line keeping an advanced use
case.

From the start I should have made this entire "feature" dependent on an Admin Setting in the database. Then there would be absolutely no changes for the large majority of users that would never use it.

Furthermore, to keep things even less invasive, the 2 new buttons should only be shown if the Org setting `Filter Attendance Roster By Subgroup` has been switched on? Then they only show for users who have knowingly desired it for that particular Org.

Again thanks for taking the time to address this pull request.

If these ideas make it less invasive enough to do the "IsLeader" (2nd method), then great. 

Otherwise I'll re-submit the "leader:" naming convention (1st method), with the fix mentioned above for disabling filtering on the web page version.

Just let me know which one you will accept and I'll do it, at this point I'm sure you're ready to close this and be done with it :)

davidcarroll:
Excellent. I like your ideas. Thank you for working with my pushback. 

When you say Admin setting, I assume you mean Admin role. Settings affect the whole database, Roles affect the user. 

Let's proceed with making these ideas available to test. I feel confident now that that will pass @karenworrell 's muster. Once she gives the go ahead, you can issue your final pull request and I will merge and publish. Also. be sure to rebase your code on to our latest master, no need to squash unless you want to.

rastographics:
> Excellent. I like your ideas. Thank you for working with my pushback.

Great! and just to be sure, are you referring to using the "naming convention" method or the "hidden buttons" method?

> When you say Admin setting, I assume you mean Admin role. Settings affect the whole database, Roles affect the user

Sorry I think I misworded what I meant. I was referring to the /Settings/ page where settings like "TransactionGateway" and "WorshipId" are kept. It seemed like that's where you like to put the flags to enable "optional" features on a "per church" level. Is this the appropriate place?

davidcarroll:
I like hiding the action items except where indicated. That solves the part about the UI exposing things that people will misinterpret, thus causing more support requests.

No need for a Setting, the checkbox on the org will do. If it becomes a problem, we will add a setting then.

rastographics:
Excellent thanks again

hkouns:
Gents, I have been following this exchange with interest.  Let me add that I am excited about the possibility of using this feature for tracking attendance in Worship Services.  I think this will be useful for churches that may not have extensive small groups in place yet or for small-medium size churches that want to do a better job tracking attendance on a weekly basis.  

If a church does not have high percentage of their attendees in a small group, the queries and metrics in BVCMS for measuring absences do not work well.  This leads to a lot of false "John Doe has not visited in 3 months" when they actually have...but they have only been in worship service.  And like many things... if the people getting the attendance/absents report see too many "False" alerts, then they think the software is "just crying wolf" ;-)

We find that folks tend to be creatures of habit and sit in the same general area.  So we have tossed around the idea over the years of assigning some folks to monitor specific areas.... jotting down the names they know...and making sure they introduce themselves to any guests and encourage them to fill out a Welcome Card.   I could see this new feature being used to potentially supplement that effort.

rastographics:
@hkouns Great Idea! 
As mentioned in the Notes of my original comment at the top, I am hoping to use this for our Worship Services too. I like your idea of having people assigned to "areas" to catch those not in any small group. 
I'm going to suggest to our team that we make a sub-group for "Not in Small Group" and assign some leaders to it so we can do exactly what you are saying.

rastographics:
I'm going to have to issue a new pull request, I messed things up with git when trying to rebase and had to use a fresh fork.

davidcarroll:
This is a good article on rebasing before pull-requests.

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

rastographics:
Closing and making [new Pull Request](https://github.com/bvcms/bvcms/pull/110) because of git issues

PR #113, opened 2016-04-26
Fix bug: Creating or Renaming a Sub-group by pressing Enter...

rastographics:
…does not post correctly (uses regular POST instead of jquery ajax post)
Also this PR enables input box to receive focus when the renaming sub-group modal opens.

The code change is simply using the jquery form "submit" event instead of button click event, so it catches the form whether it is submitted by clicking button, or submitted by hitting enter key while an input in the form is active.


rastographics:
Changes live on http://dev.nwbbc.org:44401
user: admin
password: touchpoint.1

davidcarroll:
I'll get this pulled in tonight.

davidcarroll:
This is published to production. Thank you @rastographics

PR #130, opened 2016-08-09
Ensure SGF handles various extra value types

drmohundro:
We inadvertently were only supporting text extra value types in the Small Group Finder! This should fix that - it just coalesces between whichever value is set which is easier in this case as the front end just needs string values.


davidcarroll:
Does the Extra Value search on the OrgSearch page work with the multiple ev types?

drmohundro:
Looking at https://github.com/clearfunction/bvcms/blob/master/CmsWeb/Areas/Org/Models/OrgSearchModel.cs#L281-L284 - my guess is that it currently is only handling `StrValue` or `Data`. I'll get a PR for that, too.

PR #133, opened 2016-08-12
SGF fixes related to extra values with spaces

drmohundro:
The small group finder had issues when extra values were named with spaces in them - the filter didn't see them and wouldn't show matches. In addition, the replacements wouldn't match either so they didn't work.

The other minor thing is a new admin setting called `UX-HideExtraValueEditForOrgLeaderOnly` that will still let org leaders edit individual extra values, *but* it will prevent them from creating new ones.

### DIFF
--- a/CmsWeb/Areas/Org/Views/Org/CommunityGroup/CommunityGroup.cshtml
+++ b/CmsWeb/Areas/Org/Views/Org/CommunityGroup/CommunityGroup.cshtml
@@ -52,10 +52,15 @@
 </form>
     @helper ExtraDataFormAction(ExtraValueModel evmodel, bool xs = false)
     {
+        var canAddExtraValues = User.IsInRole("Admin") ||
+            (User.IsInRole("OrgLeadersOnly")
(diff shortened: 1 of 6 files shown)


drmohundro:
@davidcarroll just an FYI, if you pull up the comparison at https://github.com/bvcms/bvcms/pull/133/files?w=1 instead (note the `?w=1`) the diff will be a lot more reasonable. The `SmallGroupFinderModel.cs` had a lot of tabs and spaces intermixed and the alignment wasn't quite right.

davidcarroll:
@mo I just published to production and the Ministry tab is gone for all users!

davidcarroll:
I just commented out your code to get it working back like it was

drmohundro:
@davidcarroll yikes - okay. I've got a similar change about to go out so I'll look over that again and submit another PR. My apologies!

davidcarroll:
I am getting a lot of grief for this. I was hoping we would not have to, but I think we are going to have to have you publish to .work, have redeemer test, and then do the pull-request.

drmohundro:
👍 understood. By the way, I'm hoping to there will be only one more pull request for any additional last minute changes.

PR #134, opened 2016-08-12
Minor hiding features for org leaders, customizable filtering of which groups are shown on SGF

drmohundro:
@davidcarroll FYI, Redeemer has tested these changes on the staging site.

The customizable filtering of organizations on the small group finder is now based on new settings:

```
<SGFSetting name="ShowOnlyName" value="NAME_TO_FILTER_ON" />
<SGFSetting name="ShowOnlyValue" value="VALUE_TO_FILTER_ON" />
```

Basically, it looks for an org extra value named "NAME_TO_FILTER_ON" with a value of "VALUE_TO_FILTER_ON" before it returns it.

In addition, this includes fixes for the ministry tab hiding. I used the same pattern for some other minor admin settings to optionally hide some settings from org leaders.

These settings are:
* `UX-HideEmailDetailsForOrgLeaders`
* `UX-HideQueriesForOrgLeaders`

As I said, this also fixes the prior bug with `UX-HideMinistryTabForOrgLeaders`. The basic boolean pattern I'm now using to ensure that we don't inadvertently hide tabs from all access users is:

```csharp
// default to false
var showTab = false;

// see if they're org leaders...
if (User.IsInRole("OrgLeadersOnly"))
{
    // then only show it if the setting is not true (defaults to false)
    showTab = !DbUtil.Db.Setting("UX-TabSetting", false);
}
// otherwise if they're access role
else if (User.IsInRole("Access"))
{
    // they just get it
    showTab = true;
}
```

### DIFF
--- a/CmsWeb/Areas/People/Views/Person/Emails/Emails.cshtml
+++ b/CmsWeb/Areas/People/Views/Person/Emails/Emails.cshtml
@@ -1,8 +1,18 @@
 ﻿@model CmsWeb.Areas.People.Models.EmailModel
+@using CmsData
 @using CmsWeb.Areas.People.Models
 @using UtilityExtensions
 @{
     var access = User.IsInRole("Access");
+    var showEmailDetails = false;
+    if (User.IsInRole("OrgLeadersOnly"))
+    {
+        showEmailDetails = !DbUtil.Db.Setting("UX-HideEmailDetailsForOrgLeaders", false);
+    }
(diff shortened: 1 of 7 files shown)


davidcarroll:
Thanks, @drmohundro, that logic for showtab is very readable. Those complex, combined cases are inherently unreadable. Sometimes you just have to do the if-else cascade to understand it yourself!

PR #149, opened 2016-09-26
Fix responsive issue with resources sub tabs

drmohundro:
We missed the dropdown that shows up on mobile responsive views (the whole "visible-xs-block" stuff). As a result, the Resources sub-tabs didn't show up on smaller screens. This fixes it.


davidcarroll:
This has been published

PR #151, opened 2016-10-25
Latest Redeemer retainer requests

drmohundro:
This PR includes the following:

* Resources can now be tied to multiple organizations and/or organization types
* SGF can now have a customizable subtitle
* Custom auto promote roles
* Some basic email customizations

FYI, the below SQL will need to be run to migrate from the old Resource columns to the new Resource columns. I can handle this once it is deployed, though.

```sql
/*
migrate from single column to join table for resources
*/

;WITH ResourcesWithOrgs AS (
	SELECT *
	FROM dbo.resource
	WHERE OrganizationId IS NOT NULL
)
INSERT INTO dbo.ResourceOrganization
SELECT ResourceId, OrganizationId
FROM ResourcesWithOrgs

;WITH ResourcesWithOrgTypes AS (
	SELECT *
	FROM dbo.resource
	WHERE OrganizationTypeId IS NOT NULL
)
INSERT INTO dbo.ResourceOrganizationType
SELECT ResourceId, OrganizationTypeId
FROM ResourcesWithOrgTypes
```

### DIFF
--- a/CmsData/Extensions/User.cs
+++ b/CmsData/Extensions/User.cs
@@ -64,7 +64,7 @@ namespace CmsData
                 UserRoles.Add(new UserRole {Role = role});
             }
         }
-        public void AddRoles(CMSDataContext db, string[] value)
+        public void AddRoles(CMSDataContext db, params string[] value)
         {
             var q = from s in value
                 join r in UserRoles on s equals r.Role.RoleName into g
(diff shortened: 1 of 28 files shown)


davidcarroll:
Looks like some fundamental things are getting changed, not just Redeemer specific. Email Controller, AddRoles, AutoPromotion, etc. What about testing to make sure there are no breaking changes?

drmohundro:
Sorry, should have clarified but Redeemer has tested/verified this in staging.

davidcarroll:
I have this published to our .work site so we can test email and roles. Once Karen is done with that, probably some time tomorrow, I'll push to production assuming all is well. Thanks Mo.

davidcarroll:
This was published to production last night

PR #152, opened 2016-10-31
Mail merge with Person properties

k4larson:
Merge word document template with properties from Person and return the merged document in the http response

### DIFF
--- a/CmsData/API/PythonModel/Person.cs
+++ b/CmsData/API/PythonModel/Person.cs
@@ -5,6 +5,8 @@ using System.Web;
 using CmsData.API;
 using CmsData.Codes;
 using UtilityExtensions;
+using Novacode;
+using System.Xml.Linq;
 
 namespace CmsData
 {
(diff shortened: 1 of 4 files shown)


CmsData/API/PythonModel/Person.cs
```diff
@@ -334,5 +336,110 @@ public void UpdateElectronicStatement(object query, bool tf)
...
+        /// and returning the merged document in the http response
+        /// </summary>
+        /// <param name="pid">The person ID to get the properties from</param>
+        /// <param name="templateName">The full path to the Word document template</param>
+        /// <param name="filePrefix">The prefix of the filename to use in the http response</param>
+        public void MergeDocument(object pid, string templateName, string filePrefix)
+        {
+            Person person = db.LoadPersonById(pid.ToInt());
+
+            DocX document = DocX.Load(templateName);
```
davidcarroll: Where is this path going to be located? It can't reside on the server's file system nor on the client's filesystem. It would be need to be uploaded to the cloud somewhere.

CmsData/API/PythonModel/Person.cs
```diff
@@ -334,5 +336,110 @@ public void UpdateElectronicStatement(object query, bool tf)
...
+            XElement root = document.Xml;
+            Merge(root, person);
```
davidcarroll: It looks like DocX is only being used to load the document, and then you manipulate the XML directly. Are there not any higher level functions in DocX that would do this?

davidcarroll:
See my comments in your code.

CmsData/API/PythonModel/Person.cs
k4larson: Ideally I would like to see the documents uploaded to Special Content, but I realize that's a bigger discussion.  I have been trying to get it to work reading from OneDrive, but so far no luck with that.  Would it not work to reside on the user's filesystem, until we can come up with a better solution?

CmsData/API/PythonModel/Person.cs
k4larson: Yes, DocX is pretty much just being used to load the document and traverse through the XML elements.  I couldn't find any merge capabilities in DocX, or any way to find all merge fields either.

davidcarroll: I'm OK with manipulating the XML as long as it is bullet proof. What is the chevron you speak of?

k4larson: I've done quite a bit of testing, but will do some more to make sure it doesn't fail.  The chevrons delimit the merge fields, if from Word you do Insert/Field/Mergefield.  It looks like this: «FirstName», where the opening chevron is a char 171 and the closing one is a char 187.

davidcarroll: Take a look at http://cathalscorner.blogspot.com/2009/02/docx-net-library-for-manipulating-word.html

Particularly the section on using Custom Properties to do trivial replacements with DocX (no direct XML manipulation).

davidcarroll: Yes, there is a Text property on a DocX document that would allow for finding all the potential replacment codes. In fact, I think using our EmailReplacement codes is the way to go instead of using the Person properties. Those are well documented and much more powerful than just people properties. Also, I think this needs to work from the anyplace the BlueToolbar is found so that you can get an entire document of letters for a group of people in the search, not just a one person at a time download. This all means that Python is not the place to do this. So the only sticking point is how to store and refer to the template. Special Content might be the place, but as of now, that is all pure text documents. 

I think it is going to need to be an upload like an Image. The best pattern for that is on the Person page, Profile > Documents. It would just be a new tab in the Special Content area. You could work on that piece.

The EmailReplacements (aka DocXReplacements) will be easy. I'll take care of that piece.

The only remaining piece is how to get it into the bluetoolbar and allow for multiple types of documents. And then actually executing the query (like a report) then downloading the results with all the baked docx files for each person merged together in to a single docx download.

PR #158, opened 2016-11-29
Code Reconstruction

joxteodocio:
Reconstructed code block using Pull Request #149:


joxteodocio:
Restored behaviour of the Resources tab:
  - when the screen dimension shrinks to mobile device scale, the tabs are converted to a pull-down list, instead.

PR #160, opened 2016-12-14
Scheduled emails

k4larson:
Update to the Python Email class to allow python to create emails that will be sent at a scheduled time.  For example, Faith Community Church wants to use this to send an automated follow up email following registrations.  The OnRegister script in the registration will create the email, scheduled at a later time so that it doesn't appear automated.


davidcarroll:
Thanks, Keith, looks pretty straight forward. I'll pull it this weekend.

davidcarroll:
This is in production now.

PR #161, opened 2016-12-19
Added ability to send SMS text messages from Python scripts

k4larson:
The younger generation sees email as old-school, so the youth ministries at Faith Community Church would like to communicate via text messages instead.  The first use will be a confirmation sent from an OnRegister python script.


davidcarroll:
@k4larson I have published this to our testing environment just to make sure it will not break anything exisiting. I'll notify you once I have published to production so you can test the Python functionality.

davidcarroll:
@k4larson This has been published to production. Please test and let us know how it is working. Thank you.

PR #163, opened 2017-01-24
CustomAccessRoles - refactoring of admin settings to special content

drmohundro:
Per my earlier email, this pull request covers the clean up / refactoring from the existing admin settings over to use the new `CustomAccessRoles.xml`. The format for the file is something like:

```xml
<?xml version="1.0" encoding="utf-8"?>
<roles>
    <role name="OrgLeadersOnly">
        <settings>
            <setting name="DisableHomePage" value="true" />
            <setting name="DisablePersonLinks" value="true" />
            <setting name="HideEmailDetails" value="true" />
            <setting name="HideExtraValueEdit" value="true" />
            <setting name="HideGuestsOrgMembers" value="true" />
            <setting name="HideInactiveOrgMembers" value="true" />
            <setting name="HidePendingOrgMembers" value="true" />
            <setting name="HideMinistryTab" value="true" />
            <setting name="HideNavTabs" value="true" />
            <setting name="HideQueries" value="true" />
            <setting name="LeadersCanAlwaysEditOrgContent" value="true" />
            <setting name="LimitToolbar" value="true" />
            <setting name="LimitedSearchPerson" value="true" />
            <setting name="CanEditCGInfoEVs" value="true" />
            <setting name="EditMemberData" value="true" />
            <setting name="OrgMembersDropAdd" value="true" />
            <setting name="OtherGroupsContentOnly" value="true" />
            <setting name="Person-ShowBlueToolbar" value="false" />
            <setting name="Organization-ShowSettingsTab" value="false" />
            <setting name="Organization-ShowBlueToolbar" value="true" />
            <setting name="Organization-ShowBlueToolbarFullEmailMenu" value="false" />
            <setting name="Organization-ShowBlueToolbarEmailMembers" value="true" />
            <setting name="Organization-ShowBlueToolbarEmailProspects" value="true" />
            <setting name="Organization-ShowBlueToolbarEmailMembersAndProspects" value="true" />
            <setting name="Organization-ShowBlueToolbarExportMenu" value="false" />
            <setting name="Organization-ShowBlueToolbarCustomReportsMenu" value="false" />
            <setting name="Organization-ShowBlueToolbarAdminGearMenu" value="true" />
            <setting name="Organization-ShowBlueToolbarMembersOnlyPage" value="true" />
            <setting name="Organization-ShowBlueToolbarVolunteerCalendar" value="true" />
            <setting name="Organization-ShowAddress" value="false" />
            <setting name="Organization-ShowOptionsMenu" value="false" />
            <setting name="Organization-ShowFiltersBar" value="false" />
            <setting name="Organization-CollapseOrgDetails" value="true" />
            <setting name="Organization-ShowBirthday" value="false" />
            <setting name="Organization-ShowAddress" value="false" />
            <setting name="Organization-ShowCreateNewMeeting" value="false" />
            <setting name="Organization-ShowDeleteMeeting" value="false" />
            <setting name="Organization-ShowTagButtons" value="false" />
            <setting name="Meeting-HyperlinkNames" value="false" />
            <setting name="Meeting-ShowAddGuest" value="true" />
            <setting name="Meeting-AllowEditDescription" value="false" />
            <setting name="Meeting-ShowExtraValuesBox" value="false" />
            <setting name="Meeting-ShowWandTargetBox" value="false" />
            <setting name="Meeting-ShowBlueToolbar" value="false" />
            <setting name="Meeting-ShowBlueToolbarIpadAttendance" value="false" />
            <setting name="Meeting-ShowBlueToolbarRollsheet" value="false" />
            <setting name="Meeting-EnableEditByDefault" value="true" />
            <setting name="Meeting-ShowEnableBox" value="false" />
            <setting name="Meeting-ShowShowBox" value="false" />
            <setting name="Meeting-ShowAttendType" value="false" />
            <setting name="Meeting-ShowOtherAttend" value="false" />
            <setting name="Meeting-ShowCurrentMemberType" value="false" />
        </settings>
    </role>
</roles>
```

We also added a new C# class to help with parsing/checking the settings. Usage would be like:

```csharp
RoleChecker.HasSetting(SettingName.CanEditCGInfoEVs, false)
```

Finally, Kayan has already signed off and tested this on staging (pointed to the staging database).

Thanks!

### DIFF
--- a/CmsData/Classes/ExtraValues/Value.cs
+++ b/CmsData/Classes/ExtraValues/Value.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Web;
 using System.Xml.Serialization;
+using CmsData.Classes.RoleChecker;
 using UtilityExtensions;
 
 namespace CmsData.ExtraValue
@@ -48,7 +49,7 @@ namespace CmsData.ExtraValue
                 if (user.IsInRole("Edit"))
                     return true;
 
-                if (user.IsInRole("OrgLeadersOnly") && DbUtil.Db.Setting("UX-OrgLeadersOnlyCanEditCGInfoEVs"))
+                if (RoleChecker.HasSetting(SettingName.CanEditCGInfoEVs, false))
                 {
                     if (string.IsNullOrEmpty(EditableRoles))
                         return true;
--- a/CmsData/Classes/RoleChecker/RoleChecker.cs
+++ b/CmsData/Classes/RoleChecker/RoleChecker.cs
@@ -1,31 +1,35 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
+﻿using System.IO;
 using System.Web;
 using System.Xml.Linq;
-using System.Xml.Serialization;
-using IronPython.Modules;
 
 namespace CmsData.Classes.RoleChecker
 {
     public static class RoleChecker
     {
-        private static XDocument xdoc
+        private static XDocument Xdoc
         {
             get
             {
+                const string CustomAccessRolesKey = "CustomAccessRoles.xml";
                 try
                 {
-                    return XDocument.Load(new StringReader(DbUtil.Content("CustomAccessRoles.xml", "")));
+                    var doc = HttpContext.Current.Items[CustomAccessRolesKey] as XDocument;
+                    if (doc == null)
+                    {
+                        doc = XDocument.Load(new StringReader(DbUtil.Content(CustomAccessRolesKey, "")));
+                        HttpContext.Current.Items[CustomAccessRolesKey] = doc;
+                    }
+                    return doc;
                 }
                 catch
                 {
-                    return new XDocument();
+                    var doc = new XDocument();
+                    HttpContext.Current.Items[CustomAccessRolesKey] = doc;
+                    return doc;
                 }
             }
         }
-        private static XElement roles => xdoc.Element("roles");
+        private static XElement Roles => Xdoc.Element("roles");
 
         private static XElement Settings(XElement role)
         {
@@ -34,6 +38,8 @@ namespace CmsData.Classes.RoleChecker
 
         public static bool HasSetting(SettingName setting, bool defaultValue)
         {
+            var roles = Roles;
+
             if (roles != null)
             {
                 foreach (var r in roles.Elements("role"))
--- a/CmsData/Classes/RoleChecker/RoleChecker.cs
+++ b/CmsData/Classes/RoleChecker/RoleChecker.cs
@@ -67,6 +67,7 @@ namespace CmsData.Classes.RoleChecker
 
     public enum SettingName
     {
+        // ReSharper disable InconsistentNaming
         Person_ShowBlueToolbar,
         Organization_ShowSettingsTab,
         Organization_ShowBlueToolbar,
@@ -102,7 +103,6 @@ namespace CmsData.Classes.RoleChecker
         Meeting_ShowAttendType,
         Meeting_ShowOtherAttend,
         Meeting_ShowCurrentMemberType,
-        AutoOrgLeaderPromoteCustom,
         AutoOrgLeaderPromotion,
         DisableHomePage,
         DisablePersonLinks,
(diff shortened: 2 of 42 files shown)


davidcarroll:
Mo, is there a need to migrate settings to this new format? We have a few church already using some of these settings.

davidcarroll:
How does this work with multiple databases in the same instance of the application? You are using a static class to cache the Roles and Settings pulled from the current database. For the current Settings, we use the the HttpRuntime.Cache and distinguish settings for different databases with the Host header.  The cache only lasts 15 seconds and will be duplicated for each web server.

See https://github.com/bvcms/bvcms/blob/master/CmsData/DbUtil/Session2.cs#L120-L131

drmohundro:
Good questions...

To answer your first question, when we had previously discussed this, I think you had said only one other church was using it (the one that Heath works with I think), so my understanding was that it wouldn't be a big deal. I may have misunderstood, though.

Regarding your second question, great thought regarding static. We discussed that as well, but we're not actually doing any caching. The static usage in `RoleChecker` is 100% methods versus storing anything. For example, the private `xdoc` property is simply a getter that returns the contents of the Special Content file - it doesn't actually store anything. Does that make sense?

drmohundro:
The reason roles are there is that the settings aren't global for all users, they only apply for users like OrgLeadersOnly. Redeemer wanted to have the flexibility to 1) have a single place to define role-specific settings versus individual admin settings and 2) to be able to pick and choose those settings for other types of roles.

Would you like us to look into adding the same caching that the admin settings uses?

davidcarroll:
It looks like you have about 50 settings you are going to be examining during the course of an Http Request and Response. If all settings were examined on a particular page, then you would be parsing the XML 100 times because you refer to the roles getter twice in HasSetting. That just does not seem good to me. Does it to you? I don't care about the caching so much across Requests, but during the request, I just don't understand why you don't put this in some sort of dictionary structure and put it in the HttpContext Items property for the life of the Request so you only read and parse the xDocument once.  

What if a setting is needed for any role? How is that configured?

drmohundro:
Ohhh, I think you're right - I think we can fix that by just putting a local variable in `HasSetting` so that the roles property gets put into a variable. Alternatively, we could also put this in HttpContext so that the XML is only loaded once per request. Good catch!

Any preference on local variable for roles versus storing in HttpContext?

Also, regarding something that is needed for _all_ roles, we just suggested to Redeemer to leave those in the existing global admin settings.

davidcarroll:
I don't think a local variable would help since this is a static class with static methods. It would be useful for a single call to HasSetting but since that can be called multiple times so it would just reduce the number of parsings from 100 to 50. It should be at the Request level which would be the Items property of the HttpContext.

drmohundro:
Got it, we'll get this loaded in HttpContext - I'll update the PR once that is in place.

drmohundro:
@davidcarroll I just pushed the `HttpContext` caching - again, good catch there!

davidcarroll:
Thank you @drmohundro

drmohundro:
@davidcarroll just checking in on this... was there anything else that you saw that you had concerns about? Thanks!

davidcarroll:
I just need some time to pull the request down and run a few tests. Been traveling and swamped with work. I'll get it done before the week's out.

drmohundro:
Gotcha, thanks David!

davidcarroll:
@drmohundro I am about to publish to `.work` to do a little bit of final testing. Then I'll publish to production sometime later today. If this breaks @hkouns database, I'll let him contact you directly. If you want to be pro-active and contact him, that would be good. He is just using the resources feature by the way. I don't know if that will affect him or not.

drmohundro:
Thanks @davidcarroll! I'll shoot an email to @hkouns as well, but this doesn't affect Resources in any way, so he should be good.

davidcarroll:
This is now in production. Thank you @drmohundro

hkouns:
I tested Resources after it was published to production and everything worked fine on our database.  @drmohundro  thanks for sending me the email as well.  I am going to try to setup small groups later this spring, so I may reach out to you for additional guidance at that time.

PR #165, opened 2017-02-15
Email prospect fix

drmohundro:
A bug was found related to the "Only Prospects" email feature (and the members and prospects list). First off, there is a recipients list that is stored off in hidden fields. It also then uses a hidden field for "only prospects" so that the Emailer doesn't load all recipients up from an org and only relies on the recipient list (which populated the `EmailQueueTos` list).

This was tested on staging and signed off by Redeemer.

### DIFF
--- a/CmsData/Email/Emailer.cs
+++ b/CmsData/Email/Emailer.cs
@@ -495,7 +495,7 @@ namespace CmsData
             return aa;
         }
 
-        public void SendPeopleEmail(int queueid)
+        public void SendPeopleEmail(int queueid, bool onlyProspects = false)
         {
             var emailqueue = EmailQueues.Single(ee => ee.Id == queueid);
             var from = Util.FirstAddress(emailqueue.FromAddr, emailqueue.FromName);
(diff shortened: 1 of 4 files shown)


davidcarroll:
I published this to .work so we can test

davidcarroll:
This has been published to production.

drmohundro:
Thanks David!

davidcarroll:
@drmohundro there is a serious error in this code at https://github.com/bvcms/bvcms/blob/master/CmsWeb/Areas/Main/Views/Email/Index.cshtml#L83
We are getting tons of null reference errors.
I am going to make a quick fix, but I need you to review your code to see how this happened since I am not familiar with what you are trying to do here.

drmohundro:
That looks correct to me - what my original code was trying to do was to associate what the users sees with what is actually sent. The "only prospects" thing went in last October I think, but it only _displayed_ the correct recipients. The back end code still goes and re-queries the organization for its members meaning the email was entirely wrong for Redeemer. We both thought it worked because it displayed the correct data. In my head, I though that I initialized the RecipientIds to a new List at the beginning but I clearly didn't do that. My apologies on missing this.

PR #166, opened 2017-02-24
Added Alt Name to all Check-In functionality

steven-bvcms:
Added Father and Mother name to Add/Edit Person
Misc other updates


davidcarroll:
This is on .work now

PR #168, opened 2017-03-31
Attempt to fix perf issue related to member only page

drmohundro:
I couldn't use `CurrOrgMembers` as is because it also returns prospects and the member only page directory only needs to show members/leaders. I ended up building a LINQ query instead... it seems to perform a lot better now, though, because it is a single query instead of an N+1 issue.


davidcarroll:
This is published to production

PR #176, opened 2017-06-21
Fix Resource Sub-Tabs on Mobile (#93)

drmohundro:
Good afternoon David - we found a minor bug with resource sub-tabs _specifically on mobile views_. This fix adds the deep link change so that mobile resource sub tabs work as well.

For a little bit more clarification, this fix is only related to the Resource tab on the Profile page so it will only affect churches that have the Resources feature enabled and are viewing the tab on mobile.

Thanks!


drmohundro:
@davidcarroll FYI, we just pushed another commit that should fix the cookie tab issue (where tabs weren't being remembered).

PR #184, opened 2017-09-15
Add extra value support to contacts

drmohundro:
This pull request adds extra value support to the contacts. The backing table is `ContactExtra`, which was recently added. One distinction with contacts is that the location is based on the ministry, contact type and contact reason.

Finally, this feature is hidden behind an admin setting named `Feature-ContactExtra`.


davidcarroll:
I'll publish Sunday evening.

PR #205, opened 2018-01-07
Update and rename ReadMe.md to README.md

charleshkang:
As someone who is new to Windows development and coming from a background in mobile/iOS development, I found the initial dev environment setup to be harder than it needed to be. The general steps were there but small yet essential steps seemed to be missing, at least when I tried setting it up.

Through talking to David, it looks like the majority of the language was updated already. I've just gone ahead and fixed some typos and fixed the code snippets to be more clear. I also renamed the file to be all-caps to follow normal conventions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ BVCMS is maintained by TouchPointSoftware, LLC.
 ### Install Development Tools and required libraries
 
 1. **[Visual Studio Community 2017][IDE]**
-    Be sure to install the ASP.NET and web development Workload option.
-    And the Python development if you want to debug python scripts.
+    - Be sure to install the ASP.NET and web development Workload option.
+    - Install the Python development option if you want to debug Python scripts.
     
 1. **[SQL Server Express 2016][SQL]**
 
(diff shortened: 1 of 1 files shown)


davidcarroll:
Hi Charles, thanks for the edits to the README file. That gets treated as a stepchild sometimes and does not get the attention it deserves. I will say, that we have had many (probably close to a hundred), people install BVCMS on their on machines over the past 9 years since we have been working on this project. A few had trouble, but those were inexperienced, non-developers. In fact, we have used this as a test -- installing the source code, building it, and making a small change and finally issuing a pull request -- for developers applying for a position. It has worked well. But I can understand if you are not familiar with the Microsoft / SQL Server stack that it could trip you up.  I created a video for you to watch https://youtu.be/xMh1k79HDdk which shows me doing a clean install. Sometimes that is the only way I can find out if things may have changed to where it needs to be updated. Indeed they have. But it sounds like you have it up and running now. Blessings. David Carroll

charleshkang:
It was definitely a combination of me being very new to SQL as well that made this more difficult, but I appreciate the video you made for me!

PR #224, opened 2018-07-19
Send group leader attendance messages

brondavies:
These are changes to support the group leader attendance message feature
group-attendance-link-setting

### DIFF
--- a/CmsData/Classes/Twilio/TwilioHelper.cs
+++ b/CmsData/Classes/Twilio/TwilioHelper.cs
@@ -93,6 +93,28 @@ namespace CmsData.Classes.Twilio
             ExecuteCmsTwilio(list.Id);
         }
 
+        public static bool SendSMS(CMSDataContext db, string toNumber, string message)
+        {
+            bool success = false;
+            string sSID = GetSid(db);
+            string sToken = GetToken(db);
+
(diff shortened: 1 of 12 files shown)


brondavies:
Need one more review @aternex - ignore whitespace 😬

CmsData/Classes/Twilio/TwilioHelper.cs
```diff
@@ -93,6 +93,28 @@ where p.ReceiveSMS
...
+                SMSNumber smsNumber = db.SMSNumbers.FirstOrDefault();
+                if (smsNumber != null)
+                {
+                    var response = SendSmsInternal(sSID, sToken, smsNumber.Number, toNumber, message);
+                    success = new[] {
+                        MessageResource.StatusEnum.Accepted,
+                        MessageResource.StatusEnum.Queued,
+                        MessageResource.StatusEnum.Sending
+                    }.Contains(response.Status);
+                }
```
aternex: 😄  i wish c# had an IN statement

brondavies: I would have used a `switch/case` but these enums aren't constants for whatever reason.  This syntax ends up being more concise anyway.

PR #231, opened 2018-08-01
remove people id from header in contribution statement

timdavila:


timdavila:
I removed the people id and campus id from header and people id from footer.

brondavies:
We want to keep the people id in the footer @timdavila . They were just concerned about the one on top and the one in the footer is sometimes used to help identify the people record.

PR #238, opened 2018-08-08
Fix Condition ContributionAmountSinceSetting

davidcarroll:
This should fix the problem with MountPisgah Recent Donors not reconciling.
It was a bug in the regular expression to split the values in the Setting Name (with start date).
They have a second parameter to fetch the fundid separated by a comma.
`BridgeSinceDate,BridgeFundId`
Both of these names are for settings in their database.
The second name was returning an empty fundid because of the regex bug.

Trello: Bridge+ Discrepancy when running script reports https://trello.com/c/O8rvfZmp
Please let me know when this is published so that I can let Jeannie know.

### DIFF
--- a/CmsData/QueryBuilder/Expressions/Contributions.cs
+++ b/CmsData/QueryBuilder/Expressions/Contributions.cs
@@ -302,7 +302,7 @@ namespace CmsData
         }
         internal Expression ContributionAmountSinceSetting()
         {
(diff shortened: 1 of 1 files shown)


davidcarroll:
This fix will need to be published to both the Web servers and to the Batch server because it runs during the MorningBatch process every morning.

PR #248, opened 2018-08-22
Austin more scoped finance reports

aternex:
Current Problems:
1. Total by Funds report set does not appear in the menu.
Added menu option, before this was only viewable to the Finance role, but nothing prevented direct navigation to it.

2. The Giving tab on People records is not seen. (This was specifically requested by the client. Should see only designated fund(s) on the Giving tab; should not see non-allowed funds, nor Statements that include all funds.)
Added filters for the contributions that make up the statement and the giving tab based on role membership. The current user can view their own page and see everything or an authorized user clicking on the giving tab would see only those contributions that were to a fund they are authorized for. This has a side effect of meaning that 2 users could see different balances for a statement if they are authorized for different funds.

3. No ability to export any reports. (Takes me to the login screen.)
Corrected the roles for this controller method and you can now export.

4. The Contributions Basics Report includes non-allowed funds.
Custom report at client level. Goal is to remove custom reports from the menu if not a member of the finance role. 

5. The Monthly Giving Analysis includes non-allowed funds.
Custom report at client level. Goal is to remove custom reports from the menu if not a member of the finance role. 

6. Totals by Funds Payment Type includes non-allowed funds.
Custom report at client level. Goal is to remove custom reports from the menu if not a member of the finance role.

### DIFF
--- a/CmsWeb/Views/Shared/Menu/Admin.cshtml
+++ b/CmsWeb/Views/Shared/Menu/Admin.cshtml
@@ -74,7 +74,7 @@
             <a href="#" class="dropdown-toggle" data-toggle="dropdown">
                 <span class="visible-xs-inline"><i class="fa fa-cogs"></i>&nbsp;&nbsp;</span><span class="hidden-lg hidden-xs">Admin</span><span class="visible-lg-inline visible-xs-inline">Administration</span> <b class="caret"></b>
             </a>
-            if (admin || manageemails || managetrans || edit || design || manageResources || financedataentry)
+            if (admin || manageemails || managetrans || edit || design || manageResources || financedataentry || financeviewonly)
             {
                 <ul id="admin-menu-col-1" class="dropdown-menu dropdown-menu-large row">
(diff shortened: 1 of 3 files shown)


CmsWeb/Views/Shared/Menu/Admin.cshtml
```diff
@@ -81,6 +81,7 @@
                         <ul>
                             <li class="dropdown-header dropdown-sub-header">Contributions</li>
                             @Helper.LiAnchorLink("Deposits Totals", "/FinanceReports/DepositTotalsForDates/")
+                            @Helper.LiAnchorLink("Totals by Fund", "/FinanceReports/TotalsByFund/")
```
brondavies: Why are we adding these menu items?  Were they in the wrong place? Removed elsewhere?

brondavies: Oh, I see what you're saying in the summary ☝️ 🤔

PR #258, opened 2018-08-31
Add migration for functions used in statement generation to include t…

aternex:
…he fund description and code to toggle the statement display text based on a dbsetting


aternex:
Validated @ pr258.tpsdb.work

PR #267, opened 2018-09-12
Fix for accessing DbUtil.Db outside of an HttpContext thread

brondavies:
Fix for https://trello.com/c/TPLknRYo/5140-statements-stuck-running-and-not-generating-statements-under-administration-contributions-statements-for-all-databases

This PR attempts to correct any use, intentional or otherwise of the helper method `DbUtil.Db` so that if it is used outside of an `HttpContext` thread it will have access to a database that was created using `DbUtil.Create(host)`  This whole class really needs to be destroyed but that would impact thousands of lines of code.  Baby steps...


brondavies:
Could use another review @aternex Thanks

PR #270, opened 2018-09-17
Add the role to the finance group on useredit page

aternex:
https://trello.com/c/KTjiyHGG/5117-add-financedataentry-role-as-option-when-editing-user-roles

### DIFF
--- a/CmsData/Extensions/User.cs
+++ b/CmsData/Extensions/User.cs
@@ -1,9 +1,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using UtilityExtensions;
 using System.Web.Security;
-using Dapper;
+using UtilityExtensions;
 
 namespace CmsData
 {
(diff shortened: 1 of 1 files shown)


CmsData/Extensions/User.cs
```diff
@@ -49,55 +48,69 @@ public void SetRoles(CMSDataContext db, string[] value, bool log = true)
...
             foreach (var s in q)
             {
                 var role = db.Roles.SingleOrDefault(r => r.RoleName == s);
                 if (role == null)
+                {
                     throw new Exception($"Role {s} does not exist");
-                UserRoles.Add(new UserRole {Role = role});
+                }
+
+                UserRoles.Add(new UserRole { Role = role });
```
brondavies: Thank you! This was really confusing until I realized you fixed the formatting.

CmsData/Extensions/User.cs
```diff
@@ -164,7 +193,7 @@ public bool CanAssign(CMSDataContext db, string role)
 
         public static string[] Financial =
         {
-            "Finance", "FinanceAdmin", "ManageTransactions", "FinanceViewOnly"
+            "Finance", "FinanceAdmin", "ManageTransactions", "FinanceViewOnly", "FinanceDataEntry"
```
brondavies: When are we adding `FundManager` to this list?  Would this be a good time to add it?

aternex: Oh, I forgot we talked about that. I will add it there too

PR #274, opened 2018-09-18
Add the limiting role to funds

aternex:
Logic was changed on how funds are accessed between roles. 

https://trello.com/c/DRdbSaQk/5115-issue-with-fund-based-finance-roles-coding
https://trello.com/c/zyeqZmz6/5032-fund-based-finance-roles


brondavies:
See my comment here: https://github.com/bvcms/bvcms/pull/270#discussion_r218192073
We should add the `FundManager` role to the `Financial` roles either in this PR or the other.

PR #276, opened 2018-09-20
Add support to filter Special Content by keywords

davidcarroll:
Each file can have one or more keywords added to it via the ContentKeywords table. This table already exists but has never been used. A comma separated list of keywords can be assigned when editing the file. This only applies to HTML, Text, Sql Scripts and Python Scripts files. The Use Case is when you have multiple related files for a special project. Using this feature, you can filter out all the noise of other, unrelated files.

The filter is entered at the top of each list of files on the various tabs. The dropdown to select the keyword is only displayed if there available keywords on files. The filter is stored in a session variable and remains on until specifically removed by selecting the "(not specified)" item in the dropdown list.

When creating a new file while a  keyword filter is set, that keyword is automatically assigned.


davidcarroll:
Apparently, something happened when I merged from upstream/master. I am going to try to start over with a new fork or try to reset back to a point in time. Working from a fork is new to me.

PR #279, opened 2018-09-20
add option to delete a twilio group. this soft deletes the group by s…

timdavila:
…witching a flag in the db


timdavila:
https://trello.com/c/KljfcwO9/5135-add-an-option-for-users-to-delete-a-twilio-group

CmsWeb/Areas/Setup/Controllers/TwilioController.cs
```diff
@@ -17,13 +17,14 @@ public ActionResult Index( int activeTab = 0 )
...
-            var n = new SMSGroup();
+            var group = new SMSGroup();
 
-            n.Name = name;
-            n.Description = description;
-            n.SystemFlag = systemFlag;
+            group.Name = name;
+            group.Description = description;
+            group.SystemFlag = systemFlag;
+            group.IsDeleted = false;
```
aternex: If affects nothing, but if you wanted to save some key strokes in the future you can use the object initializer pattern if you are setting properties along with the act of creating an object. 

var group = new SMSGroup { Name = name, Description = description, SystemFlag = systemFlag, IsDeleted = false };

CmsData/Generated/SMSGroup.cs
```diff
@@ -129,37 +134,59 @@ public string Description
...
-        }
-
+		}
 
-        [Column(Name = "SystemFlag", UpdateCheck = UpdateCheck.Never, Storage = "_SystemFlag", DbType = "bit")]
-        public bool SystemFlag
-        {
-            get { return this._SystemFlag; }
+		
+		[Column(Name="is_deleted", UpdateCheck=UpdateCheck.Never, Storage="_IsDeleted", DbType="bit NOT NULL")]
```
brondavies: It's a small thing but for consistency - we like to have the column name match the property name so this should be `IsDeleted` not `is_deleted` (also in the migration script)

CmsData/Migrations/20180919_add-sms-groups-deleted-flag.sql
```diff
@@ -0,0 +1,4 @@
+﻿BEGIN
```
brondavies: Add an `IF NOT EXISTS` block to only attempt the `ALTER TABLE` if it doesn't have the column.  It makes it safe to run the script against any database where we aren't sure what state it's in - which is almost always 😂

CmsWeb/Areas/Setup/Views/Twilio/DeleteDialog.cshtml
```diff
@@ -0,0 +1,21 @@
+﻿@using CmsData
```
brondavies: Can you rename this file and it's references to `DeleteGroupDialog` so that it is more specific. Thanks.

PR #281, opened 2018-09-24
Sql migration to delete Custom Task Role.

josearmandohf:
This migration deletes role using RoleName = 'Task' from UserRole and Roles tables.

### DIFF
--- a/CmsData/Migrations/20180924_Remove_Custom_Task_Role.sql
+++ b/CmsData/Migrations/20180924_Remove_Custom_Task_Role.sql
@@ -1,0 +1,19 @@
+﻿DECLARE @RoleID INT
+DECLARE @RoleName VARCHAR(100)
+SET @RoleName = 'Task'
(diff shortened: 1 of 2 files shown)


CmsData/Migrations/20180924_Remove_Custom_Task_Role.sql
```diff
@@ -0,0 +1,19 @@
+﻿DECLARE @RoleID INT
+DECLARE @RoleName VARCHAR(100)
+SET @RoleName = 'Task'
+IF EXISTS(SELECT 1 FROM [Roles] WHERE RoleName like @RoleName )
+BEGIN
+	IF (SELECT COUNT(roleid) FROM [CMS_bellevue].[dbo].[Roles] WHERE RoleName like @RoleName) < 2
```
brondavies: There shouldn't be a database name here

josearmandohf: Thanks Bron, I have removed DB name.

PR #282, opened 2018-09-24
Add support to filter Special Content files by keyword

davidcarroll:
Each file can have one or more keywords added to it via the ContentKeywords table. This table already exists but has never been used. A comma separated list of keywords can be assigned when editing the file. This only applies to HTML, Text, Sql Scripts and Python Scripts files. The Use Case is when you have multiple related files for a special project. Using this feature, you can filter out all the noise of other, unrelated files.

The filter is entered at the top of each list of files on the various tabs. The dropdown to select the keyword is only displayed if there available keywords on files. The filter is stored in a session variable and remains on until specifically removed by selecting the "(not specified)" item in the dropdown list.

When creating a new file while a keyword filter is set, that keyword is automatically assigned.

### DIFF
--- a/CmsData/Extensions/Content.cs
+++ b/CmsData/Extensions/Content.cs
@@ -1,4 +1,6 @@
+using System.Collections.Generic;
 using UtilityExtensions;
+using System.Linq;
 
 namespace CmsData
 {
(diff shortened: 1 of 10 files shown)


davidcarroll:
FYI, turns out that on the previous PR, I was on the wrong branch (not master) when I started the original ContentLabels branch. I believe when I merged from upstream/master I ended up with a mess of commits that had nothing to do with this branch. Anyway, I deleted my fork and added my changes back manually from a local copy. That allowed me to create this new PR with a very clean commit on a fresh branch called ContentKeywordFilter. I am still learning the workflow for GitHub contributions from a fork and how to keep your fork in sync with the upstream.

davidcarroll:
Here is a video demo of the functionality of this project.

hkouns:
I love this feature!  I think I may have even submitted a feature request a long time ago for something to fix this issue.   I will use a keyword to filter all of the HTML Shells for registration pages. (and a host of other reasons)   Good Job.

brondavies:
@davidcarroll It would be nice to have email templates be able to use this feature as well.  I think the tabs for email templates and drafts should probably stay visible when filtering by keyword even though it's not supported.  What do you think?

davidcarroll:
@brondavies I can't see any reason for needing access to those two tabs when working on a project, it would be confusing. The idea was to to be able to focus on related files and eliminate all the clutter. I removed them for that reason. Easy enough to unset / reset the filter to access them if needed.

davidcarroll:
@brondavies I can do this for email templates and drafts too if you want. It just was not necessary for the purposes I had in mind. Perhaps that can be a separate project so that we can get this published.

PR #284, opened 2018-09-28
Daniel save payment box

dahnbaez:
In OneTimeGiving, and OnePageGiving, "Save Payment Information" checkbox is hidden now when payment information is already saved in DB. If User modifies this payment information checkbox shows up as unchecked forcing the user to consciously check the box if she wants to save it.


dahnbaez:
I've reverted the web.config changes, @aternex

brondavies:
At some point we'll want to apply the automatic reformatting in the web.config so that it doesn't cause us headaches down the road.

PR #286, opened 2018-10-03
Reports menu is disabled for all user with rol "OrgLeadersOnly"

dahnbaez:
All reports from the main menu are disabled for users with "OrgLeadersOnly". This type of user only can see reports from blue nav bar in Organization page that user leads.


brondavies:
@dahnbaez  Can you remove package-lock.json from this PR?

dahnbaez:
@brondavies Done

PR #288, opened 2018-10-04
Updated mobile authentication routines to include OrgLeadersOnly

steven-bvcms:
Misc other

### DIFF
--- a/CmsWeb/Areas/Public/Models/MobileAPIv2/MobileAuthentication.cs
+++ b/CmsWeb/Areas/Public/Models/MobileAPIv2/MobileAuthentication.cs
@@ -112,10 +112,18 @@ namespace CmsWeb.Areas.Public.Models.MobileAPIv2
 			}
 
 			foreach( var foundUser in userQuery.ToList() ) {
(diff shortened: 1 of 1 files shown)


CmsWeb/Areas/Public/Models/MobileAPIv2/MobileAuthentication.cs
```diff
@@ -112,11 +112,19 @@ private Error validateBasic( string value )
 			}
 
 			foreach( var foundUser in userQuery.ToList() ) {
-				if( !Membership.Provider.ValidateUser( username, password ) ) continue;
+				if( !impersonating && !Membership.Provider.ValidateUser( username, password ) ) continue;
```
brondavies: You could speed this routine up by moving the ValidateUser call outside the for loop.

steven-bvcms: I don't think it can be on the outside since it can have multiple user matches via an email address.  I have to check each of the records to see if the email/password combination works.

PR #290, opened 2018-10-04
Daniel OrgLeadersOnly Reports

dahnbaez:
The following reports are now disabled for people with "OrgLeadersOnly" role.

Statistics:
-Vital Stats
Decisions:
-Weekly decisions
-Decision Summary
Extra Value Summary:
-People
-Family
-Organization

All others reports are available for this role.


CmsWeb/Areas/Reports/Controllers/ReportsController.cs
```diff
@@ -189,6 +189,8 @@ public ActionResult Contacts(Guid? id, bool? sortAddress, string orgname)
         [HttpGet]
         public ActionResult Decisions(int? campus, string dt1, string dt2)
         {
+            if (Util2.OrgLeadersOnly)
```
brondavies: In the future, use `{}` for all logic blocks

hkouns:
I like this approach.

PR #291, opened 2018-10-08
Josearmando pushpay authentication

josearmandohf:
Changes to execute Pushpay/Complete action anonymously and redirect to TenantSite. Minor changes in config variables.


CmsWeb/Code/CmsController.cs
```diff
@@ -100,12 +100,18 @@ protected override void OnActionExecuting(ActionExecutingContext filterContext)
...
+            var contr = filterContext.RouteData.Values["Controller"].ToString();
+            var act = filterContext.RouteData.Values["Action"].ToString();
+
             if (!User.Identity.IsAuthenticated)
             {
-                var s = "/Logon?ReturnUrl=" + HttpUtility.UrlEncode(Request.RawUrl);
-                if (Request.QueryString.Count > 0)
-                    s += "&" + Request.QueryString.ToString();
-                filterContext.Result = Redirect(s);
+                if (!(contr.ToUpper() == "PUSHPAY" && act.ToUpper() == "COMPLETE"))
```
brondavies: Why isn't this code in the Pushpay controller `Complete` method?

josearmandohf: Pushpay controller inherits from CmsStaffController with RequireHttpsAttribute.
CmsStaffController overrides OnActionExecuting forwarding any not auth user to logon avoiding Pushpay/Complete to be executed anonymously, I changed it to ask if request is comming from Pushpay/Complete then avoid redirection to logon.

brondavies: Could we use a different Controller base class?

PR #295, opened 2018-10-12
change the old source field to origin, add a true source field to ind…

timdavila:
…icate where a contribution came from


timdavila:
There isn't a specific trello card for this, Steven and I discussed this change in slack when discussing keeping track of contributions that were added as part of the pushpay project. This enables keeping track of various sources, and renames the old field that tracked whether the contribution came from the app for convention.

PR #300, opened 2018-10-18
Giving dashboard

davidcarroll:
Two Python functions to help with Giving Dashboard report.
One function to help with running SQL scripts with parameters during debugging.
Update to Contributions2 function to use the Married flag on lookup.MaritalStatus.


davidcarroll:
Just wanted to make sure you saw this as a new Pull Request since it has the same name and branch as the previous #296. See previous comment.

PR #302, opened 2018-10-18
pressing enter will take you to the first item if no item has been selected from the search results

aternex:
https://trello.com/c/BQUa3E4n/5189-while-doing-a-quick-search-the-1st-found-match-should-be-opened-when-the-user-presses-enter


davidcarroll:
I am so happy to see this! This has bugged me for a very long time. Just never got around to fixing it. Thank you.

PR #303, opened 2018-10-19
Td add pushpay pk

timdavila:


timdavila:
I didn't realize that IDENTITY is not the same as a primary key in MSSQL - rookie mistake. This PR actually sets the ID as the PK

PR #307, opened 2018-10-24
fix typo in migration preventing local build

aternex:
correct bug in master preventing local development due to database creation error


aternex:
Fielder tried to build master and received an error during db creation. The error happened when running the migrations. There was a typo for an alias to a table and this PR corrects that error. I was able to build master and get the db created with this change in place.

PR #308, opened 2018-10-24
update abundant life importer to accept new file format

timdavila:
https://trello.com/c/4UzfSOIb/5200-2-kindrid-text-to-give-import-paul-cumberland-abundantlife-25985

### DIFF
--- a/CmsWeb/Areas/Finance/Models/BatchImport/AbundantLifeImporter.cs
+++ b/CmsWeb/Areas/Finance/Models/BatchImport/AbundantLifeImporter.cs
@@ -8,6 +8,7 @@
 using System;
 using System.IO;
 using CmsData;
+using CmsData.Codes;
 using LumenWorks.Framework.IO.Csv;
 using UtilityExtensions;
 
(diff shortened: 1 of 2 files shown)


CmsWeb/Areas/Finance/Models/BatchImport/AbundantLifeImporter.cs
```diff
@@ -8,41 +8,83 @@
...
-            using (var csv = new CsvReader(new StringReader(text), hasHeaders:false))
+            using (var csv = new CsvReader(new StringReader(text), true))
                 return BatchProcessAbundantLife(csv, date, fundid);
         }
 
         private static int? BatchProcessAbundantLife(CsvReader csv, DateTime date, int? fundid)
         {
             BundleHeader bh = null;
+            csv.MissingFieldAction = MissingFieldAction.ReplaceByEmpty;
+
```
aternex: Are we certain about this? If the fields are all string values, how is the code to know which field is missing? Can we end up in a situation where a field in the middle is missing and we end up putting incorrect values in the wrong columns? Often an exception can be useful and at least give the user a chance to validate they have everything correct before proceeding with an import, but on that same note... reducing user frustration is key! More or less just wanted to make a note in case it was something that wasn't considered.

timdavila:
> Can we check for any other churches that might be using the existing format? In that case we might have to create this as a new importer.

This is why I updated the abundant life importer, it should only be for them right?

CmsWeb/Areas/Finance/Models/BatchImport/AbundantLifeImporter.cs
timdavila: I had to do this to support the subtotal rows in the csv, see line 54. Is there a better way?

brondavies:
> > Can we check for any other churches that might be using the existing format? In that case we might have to create this as a new importer.
> 
> This is why I updated the abundant life importer, it should only be for them right?

Ideally, yes, but I think it's conceivable that someone else has the same bank and they use the same format.  I will check the other databases to be sure before merging

CmsWeb/Areas/Finance/Models/BatchImport/AbundantLifeImporter.cs
brondavies: It looks ok to me for this import.

brondavies:
@timdavila So, I am seeing this text as a trigger for using this importer

`text.Substring(0, Math.Min(text.Length, 20)).Contains("10444063,")`

Does that still apply with the new file format and what is that string?  An account number?

timdavila:
> @timdavila So, I am seeing this text as a trigger for using this importer
> 
> `text.Substring(0, Math.Min(text.Length, 20)).Contains("10444063,")`
> 
> Does that still apply with the new file format and what is that string? An account number?

I'm not sure what that was. From what we can tell they changed the format and I updated that conditional to match the new format.

PR #309, opened 2018-10-25
Fix org members move process to remove members from prev org attendance list

josearmandohf:
OrgMembersAsOfDate function was retrieving members without joining OrganizationMembers table, I have add that change in Migration line 38 

JOIN dbo.OrganizationMembers om ON om.PeopleId = p.PeopleId and om.OrganizationId = @orgid


CmsData/Migrations/20181025_update_OrgMembersAsOfDate_func.sql
```diff
@@ -0,0 +1,65 @@
...
+	ms.Description MemberStatus, 
+	pp.Name BFTeacher, 
+	pp.PeopleId BFTeacherId, 
+	p.Age, 
+	mt.Description MemberType, 
+	mt.Id MemberTypeId, 
+	ISNULL(ee.EnrollmentDate, ee.TransactionDate) Joined 
+FROM dbo.EnrollmentTransaction ee 
+JOIN dbo.People p ON ee.PeopleId = p.PeopleId 
+JOIN dbo.OrganizationMembers om ON om.PeopleId = p.PeopleId and om.OrganizationId = @orgid
```
josearmandohf: This is the line I added.

davidcarroll:
A previous org member will not be in the OrganizationMembers table unless he rejoined. Only in the enrollmenttransactions table. I think this change will cause problems.

josearmandohf:
I have updated the branch to update previous organization member status. Changes in function OrgMembersAsOfDate were dismissed since those are not required for the fix.

CmsWeb/Areas/Dialog/Models/MoveOrgMembersModel.cs
```diff
@@ -76,6 +76,8 @@ public static void DoMoveWork(MoveOrgMembersModel model)
                 if (oid == model.TargetId)
                     continue;
                 OrganizationMember.MoveToOrg(db, pid, oid, model.TargetId, model.MoveRegistrationData, model.ChangeMemberType == true ? model.MoveToMemberTypeId : -1);
+                //Once member has been inserted into the new Organization then update member  in Previous Organization as not enrolled.
+                db.RepairTransactions(oid);
```
davidcarroll: I don't think you should be updating the EnrollmentTransaction record, just insert a new record with TypeId=5 (Drop). But all this should be automatically handled if you use the OrganizationMember.cs Drop and Insert methods.

davidcarroll: The history of joins, changes and drops for an organization is kept in the Enrollment TransactionTable and OrgMemberAsOf works based on that structure.

PR #315, opened 2018-11-01
use gross amount instead of net

timdavila:


timdavila:
https://trello.com/c/4UzfSOIb/5200-2-kindrid-text-to-give-import-paul-cumberland-abundantlife-25985

PR #316, opened 2018-11-01
Manage Volunteers Changes

dahnbaez:
For Manage Volunteers page, (1) add confirmation button when deleting a document, and (2) create a new role to view documents but not be able to change checkboxes, etc.

### DIFF
--- a/CmsData/Migrations/20181031_add_viewvolunteerapplication_role.sql
+++ b/CmsData/Migrations/20181031_add_viewvolunteerapplication_role.sql
@@ -1,0 +1,6 @@
+IF NOT EXISTS(SELECT * FROM dbo.Roles WHERE RoleName='ViewVolunteerApplication')
+BEGIN
+    INSERT INTO dbo.Roles (RoleName, hardwired)
+    VALUES ('ViewVolunteerApplication', 1)
+END
+GO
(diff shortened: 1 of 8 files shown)


CmsWeb/Areas/People/Views/Person/Ministry/Volunteer.cshtml
```diff
@@ -31,7 +31,7 @@
     </table>
 </div>
 @FormAction(true)
-<br/>
+<br />
```
aternex: extremely nitpicking and just a note for future code...

We have already seen issues where html emails do not render properly when the tags are written like this, even though it is currently valid per w3c. There is nothing that would lead me to believe any of the current browsers out there would render this incorrectly.

brondavies:
I need to know that we've tested this on a PR test site before we deploy.

CmsData/Migrations/20181031_add_viewvolunteerapplication_role.sql
```diff
@@ -0,0 +1,6 @@
+IF NOT EXISTS(SELECT * FROM dbo.Roles WHERE RoleName='ViewVolunteerApplication')
+BEGIN
+    INSERT INTO dbo.Roles (RoleName, hardwired)
+    VALUES ('ViewVolunteerApplication', 1)
+END
+GO
```
aternex: This would need to be ran on all databases at the time of deployment?

dahnbaez: Yes. Could send an error if any DB doesn't have it.

PR #317, opened 2018-11-05
include individual (non family) contributions also

timdavila:
https://trello.com/c/eqR17pIC/5220-1-statements-kit-brunson-smihaiti-26174


timdavila:
difference between this and the old function (20181022-1_update_Contributions2_func.sql)
https://www.diffchecker.com/RMNldt5i

brondavies:
@davidcarroll This is a fix to a query you were working on recently. Let me know if this still functions the way you intended. LGTM

davidcarroll:
@brondavies @timdavila  Apparently I was thinking about how `inner join` is the same as `join` but I conflated that with `left join` same as `join` which is obviously wrong.  My bad.

PR #318, opened 2018-11-06
Revert recent changes to DynamicData class

aternex:
bugfix preventing contribution imports, revert to old code

### DIFF
--- a/CmsData/API/PythonModel/DynamicData.cs
+++ b/CmsData/API/PythonModel/DynamicData.cs
@@ -1,9 +1,9 @@
-﻿using System;
+﻿using IronPython.Runtime;
+using Newtonsoft.Json;
+using System;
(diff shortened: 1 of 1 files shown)


davidcarroll:
@brondavies Can you give me some details on which church and which script was broken? I need to be able to see that so that I can ensure the changes work. I'll need to download that church's database and get the csv file that is not working to see what was broken.

Did you publish to production? I have already released my scripts for MP and actually went over them today in a conference call. That would mean the new GivingDashboard is now broken since it depends on the new DynamicData class.

timdavila:
@davidcarroll The issue affected all churches that used the Python API to upload files. The issue was that form elements (file uploads and date fields typically) were not added to the Data property, so the form data wasn't accessible via `model.Data` as the Python scripts expect it to.

PR #319, opened 2018-11-07
Increase SMS limit to 1600 characters

dahnbaez:
Observations:

- I've found something curious. When you edit a person's cell phone with the country code prefix (ex: +1) and then you click on save, '+' symbol disappears (a trigger in DB). The thing is the DB has the number without that '+' but Twilio needs that '+' to send SMS. I modified the SMSItem creation process to add that symbol to the number that Twilio receives.


CmsData/Classes/Twilio/TwilioHelper.cs
```diff
@@ -58,7 +58,7 @@ public static void QueueSms(IQueryable<Person> q, int iSendGroupID, string sTitl
 
                 if (!string.IsNullOrEmpty(i.CellPhone))
                 {
-                    item.Number = i.CellPhone;
+                    item.Number = string.Format("+{0}",i.CellPhone);
```
brondavies: I think since we tested/used SMS only with U.S. numbers, this was working previously - without the `+` I imagine international numbers wouldn't work, so extra points for figuring that out!  But I am curious whether the number now has to have the country code in order for this to work on all numbers.  Can we test this?

brondavies: Note also that this is inside `QueueSms` and the other methods for sending a one-off SMS [SendSMS](#diff-8a9458a24c2da0af67eefb3ca1f6690dL120) aren't enforcing the same format.

CmsData/Migrations/20181106_alter_SMSList_Message_column.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE dbo.SMSList
+	DROP CONSTRAINT DF_SMSList1_slMessage
+GO
+ALTER TABLE dbo.SMSList
+	ALTER COLUMN Message VARCHAR (1600) NOT NULL
```
brondavies: Should be `NVARCHAR`

dahnbaez: Right. My bad.

CmsData/Classes/Twilio/TwilioHelper.cs
dahnbaez: All my tests with no + numbers are throwing an error from Twilio. Even those from U.S. Twilio Doc indicate that you need to use that format too. Strange...

dahnbaez: You're right, I removed that part and should work for U.S. numbers as the previous version.
For accepting international numbers, Jose and I think that would be a good idea to add a dropdown with country codes in the CellPhone form.  Also, we could add a country code column in person table so a new lookup table with that codes. Maybe an enhancement in the future?

PR #323, opened 2018-11-12
Giving dashboard additional contribution search capabilities

davidcarroll:
@brondavies it would be very helpful if you could run the migration script 20181111_alter_Contributions2SearchIds_func.sql today on MP database. This does not depend on any of the other code in this PR. This will allow me to wrap up the first phase of the project with MP right away. No one else is using this function at this point so it should innocuous.  Thank you.


brondavies:
I've run the script on MP's database @davidcarroll

davidcarroll:
Thank you!

PR #325, opened 2018-11-13
Td checkin ui refresh

timdavila:


timdavila:
Looks like this isn't just my UI updates, but the entire check in feature.
@brondavies Let me know if I should remove this PR or change the base branch to austins

PR #327, opened 2018-11-14
Add simple injector and fix many dbutil.db calls

aternex:
Very large PR, idea is to add DI and this includes converting almost all of the spots in the controllers to this format. A large majority of the changes are whitespace changes or code format changes and do not have modifications to the logic. 

A future PR should fix the models to either not have data access (preferred w/ services in the controller to build the model) and a later push to remove DbUtil.Db from the views. These are much more complicated than just getting DI into the pipeline. Another easy to adjust area would be the batch contribution importers. There is a lot of activity in these that could be adjusted quickly.


davidcarroll:
@aternex I would suggest that rather than mix code logic into all the reformatting, non-logic changes, that you instead do two projects, one to just change code formatting with no logic changes, then from there, you can have much cleaner and easy to see logic changes in future pull requests.  Honestly, I tried to find the crux of the new model and could not find it for all the code reformatting.

aternex:
I completely agree with this statement. I have an extension in visual studio that takes a predefined "template" and formats the code to match that format on save. Ultimately, the amount of changes that were made where adding the constructors to the controllers and adding the base controller to the bottom of the current controller chain. Since these files were opened, it auto formatted them on save. It looks like a lot more changes than it really is and certainly generates noise. I may see if I can separate these into different commits as no one can be expected to code review 490+ files.

-- For textual purposes --

Problem to be solved:
*Add DI to the platform to give us a solid foundation moving toward a SOLID code base for future extensions and enhancements.
*We have been seeing issues that are difficult to explain and all seem to gravitate towards some static methods in the DbUtil class that reference the HttpContext.Current. We want to simplify creation of the database object and reduce dependencies on the HttpContext directly giving us a clear cut direction for unit test and integration test frameworks.

Summary of the real changes:
*DbUtil.Db was replaced with CurrentDatabase for most controller methods. Static methods still call DbUtil.Db for now and will be modified/targeting in the future. I want to determine if the static identifier is required for current usage or if it can be removed.

Added code in CmsWeb/Lifecycle:
*a new controller that other cmscontrollers extend with the properties we use
*a new object RequestManager that serves as the container for the HttpContext, the CMSDataContext, and the CMSImageDataContext
*added the bootstrap code for simple injector to wire everything together (global.asax, app_start, etc)

Future enhancements:
*Where possible, remove data access from the model classes. Lazy loading is useful, so be mindful moving forward not to decrease performance
*Attempt to migrate business logic into service layer to promote reuse, DI will help make this cleaner at that point
*Static methods in non-static controllers... are these required to be static or can they work off of an instance, if so, use CurrentDatabase instead of DbUtil.Db
*Where possible, remove data access from the views. Try to pass complete object to the view so it can render in full (see note about lazy loading above)
*HttpContext.Current can be replaced with IRequestManager.CurrentHost or shortcut methods on the controller/service class.
________________________________

PR #329, opened 2018-11-16
Python method disable fund by PeopleId and FundId

dahnbaez:
This are the new methods 

model.DisableFund(int PeopleId, int FundId)
model.EnableFund(int PeopleId, int FundId)

Also, I edited Manage Giving View and other methods to hide disabled funds.


CmsData/API/PythonModel/ManagedGiving.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Linq;
+using UtilityExtensions;
+
+namespace CmsData
+{
+    public partial class PythonModel
+    {
+        public void EnableFund(int peopleId, int fundId)
```
brondavies: Will you combine these into one method called `SetRecurringGivingEnabled` that takes a third boolean parameter to enable or disable the RecurringAmmounts record?

PR #332, opened 2018-11-20
Fix GetContributionDetails function to display Export Donor Totals name properly

josearmandohf:
Fix GetContributionDetails function to manage null values and properly display name in Export Donor Totals report.


CmsData/Migrations/20181120_Update_GetContributionDetails_func.sql
```diff
@@ -0,0 +1,86 @@
...
+	END AS CreditGiverId2,
+
+    CASE WHEN fa.HeadOfHouseholdId = sp.PeopleId
+		THEN p.PeopleId 
+		ELSE sp.PeopleId 
+	END AS SpouseId,
+	
+    CASE WHEN fa.HeadOfHouseholdId = sp.PeopleId
+		THEN sp.Name2 
+		ELSE ISNULL(p.Name2, CONCAT(p.LastName, ', ', p.FirstName))
```
josearmandohf: Here´s the change, the record mentioned in trello card had Name2 set as null.

PR #338, opened 2018-11-30
fix statement creation for all

aternex:


aternex:
PR340 is these changes and the gcm changes from last week as one commit. The GCM was a similar issue as this but dealt more in the APIs. That PR is currently deployed at PR340 on .work. These also fix the .net framework back to 4.6.1 instead of 4.6.2 across the sln and unless there is a reason to do each PR separate for audit/tracking it might be easier to use that branch as I have already combined the 2 conflicts between this and the gcm changes.

PR #343, opened 2018-12-06
Fix error when remove enrollment hystory in Org Member dialog

dahnbaez:
Use Stored Procedure to remove enrollment history by org id and people id instead use DeleteOnSubmit


timdavila:
looks ok, but why use a stored proc instead of do it in code?

PR #357, opened 2018-12-19
Removing spaces from RoleList, consolidating references.

josearmandohf:
1) Removing spaces from RoleList to be able to match against DB role name/value.
2) Consolidating nuget references among projects.

### DIFF
--- a/CmsWeb/Areas/People/Models/Contact/ContactModel.cs
+++ b/CmsWeb/Areas/People/Models/Contact/ContactModel.cs
@@ -18,6 +18,7 @@ namespace CmsWeb.Areas.People.Models
     {
         private bool? canViewComments;
         private string _incomplete;
+        private string _limitToRole;
 
         [NoUpdate]
         public int ContactId { get; set; }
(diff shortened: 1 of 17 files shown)


CmsWeb/Areas/People/Models/Contact/ContactModel.cs
```diff
@@ -48,16 +52,16 @@ public class ContactModel : IValidatableObject
...
             }
 
             var list = roles.Select(rolename => new SelectListItem
             {
-                Value = rolename,
-                Text = rolename,
+                Value = rolename.trim(),
+                Text = rolename.trim(),
                 Selected = !string.IsNullOrWhiteSpace(LimitToRole) && LimitToRole == rolename
             }).ToList();
```
brondavies: Not sure what happens when 2 `SelectListItem`s have `Selected = true` but the next line added the (not specified) option and sets it as Selected also.  Might want to fix that logic so that it's only selected if no other role is selected.

PR #371, opened 2019-01-08
Fix birthdate allowed formats

dahnbaez:


CmsWeb/healthcheck.txt
```diff
@@ -1 +1 @@
-﻿down
\ No newline at end of file
+﻿downs
```
brondavies: This needs to be reverted

UtilityExtensions/Extensions/Dates.cs
```diff
@@ -154,8 +154,10 @@ public static bool DateValid(string dt, out DateTime dt2)
         }
         public static bool BirthDateValid(string dob, out DateTime dt2)
         {
+            string[] formats = { "MM/dd/yyyy", "MM/d/yyyy", "M/dd/yyyy", "M/d/yyyy" };
```
brondavies: this should also include `MM/d` and `MM/dd` for optional birth year.

CmsWeb/healthcheck.txt
aternex: I will undo this, I just wanted to trigger a build quickly

UtilityExtensions/Extensions/Dates.cs
brondavies: You may have to handle those as a separate logic block lo differentiate how this function handles a date with and without a year.

dahnbaez: mmm, but we need to make the birth year required, right?

UtilityExtensions/Extensions/Dates.cs
```diff
@@ -154,8 +154,10 @@ public static bool DateValid(string dt, out DateTime dt2)
...
+
             dt2 = DateTime.MinValue;
-            if (DateTime.TryParseExact(dob, "m", CultureInfo.CurrentCulture, DateTimeStyles.None, out dt2))
+            if (!DateTime.TryParseExact(dob.ToString(), formats, CultureInfo.CurrentCulture, DateTimeStyles.None, out dt2))
```
brondavies: the `!` is probably not what you want here and you also don't need the `ToString()`

PR #376, opened 2019-01-14
Update in CMSData, creating function to get people ID by Payer Key

josearmandohf:
CMS Data updates to find people by Gateway Payer Key.


CmsData/Migrations/20190114_ja_create_fnc_FindPersonByPayerKey.sql
```diff
@@ -0,0 +1,26 @@
...
+END
+GO
+
+CREATE FUNCTION [dbo].[FindPersonByPayerKey](@GatewayPayerKey nvarchar(100), @GatewayPayerValue nvarchar(max))
+RETURNS @t TABLE ( PeopleId INT)
+AS
+BEGIN
+--DECLARE @t TABLE ( PeopleId INT)
+--DECLARE @GatewayPayerKey nvarchar(100) = 'GatewayPayerKey', 
+--		@@GatewayPayerValue nvarchar(max) = 'MzpMNU9YSTIwSmt4REN5ZWluUlhSbGNvLV9lUk0',
```
brondavies: Remove commented code

CmsData/Extensions/Person.cs
```diff
@@ -2396,10 +2396,29 @@ public void UpdateElectronicStatement(CMSDataContext db, bool tf)
...
+
+            if (!String.IsNullOrEmpty(payerKey))
+            {
+                // 1.Do we have a pushpay key that matches a peopleextra
+                list = db.FindPersonByExtraValue("GatewayPayerKey", payerKey).ToList();
+            }
+            else if (!String.IsNullOrEmpty(email))
+            {
+                // 2.Do we have an email address that matches one person
+                list = db.FindPerson(first, last, null, email, phone.GetDigits()).ToList();
```
timdavila: We'll need to leave the name off of this call - FindPerson will only return a result for matching email AND name. There are instances we ran into on some clients where a nickname or maiden name was used, and ended up creating a new person because they weren't exact matches.

CmsData/Migrations/20190114_ja_create_fnc_FindPersonByPayerKey.sql
josearmandohf: Removed.

CmsData/Extensions/Person.cs
josearmandohf: @timdavila I have created a FindPersonByEmail logic to address the issue.

CmsData/Extensions/Person.cs
```diff
@@ -2396,10 +2396,29 @@ public void UpdateElectronicStatement(CMSDataContext db, bool tf)
...
+                                           string extraValue = null)
         {
             Person person = null;
-            var list = db.FindPerson(first, last, null, email, phone.GetDigits()).ToList();
+            List<View.FindPerson> list = null;
+
+            if (!String.IsNullOrEmpty(extraValue))
+            {
+                // 1.Do we have a pushpay key that matches a peopleextra
+                list = db.FindPersonByExtraValue("GatewayPayerKey", extraValue).ToList();
```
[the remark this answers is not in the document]
josearmandohf: I've updated ExtraValueField as optional parameter in FindAddPerson method.

PR #381, opened 2019-01-16
Apply migrations like the deployment script does

brondavies:


CmsData/API/PythonModel/Handlebars.cs
```diff
@@ -119,7 +119,7 @@ public static void RegisterHelpers(CMSDataContext db, PythonModel pm = null)
                 var ntoks = args.Length > 2 ? args[2].ToInt() : 2;
                 var sep = args.Length > 3 ? args[3].ToString() : " ";
                 var a = s.SplitStr(sep, ntoks);
-                writer.Write(a[n].trim());
+                writer.Write(a[n]?.Trim() ?? "");
```
aternex: 💯

PR #396, opened 2019-01-30
Update FindPersonByExtraValue sql function.

josearmandohf:
Function updates to search any column field type.


brondavies:
Closing since it doesn't look like we'll need this

PR #399, opened 2019-02-08
use the existing db context when running python callscript

timdavila:


brondavies:
@davidcarroll The problem we are fixing is that the setup of the context is lost since it doesn't have the same properties of the previous context, particularly during morning batch where there is no HttpContext.  This also reduces the number of open connections to the database server by re-using the context.  Under what circumstances would this break in the way you've described?

davidcarroll:
Thanks for the information. 

I remember making a conscious decision to create a new dbcontext. I believe the scenario I was thinking of is that I wanted calling the script to be like calling a webservice which would create a new dbcontext in that case. Some of these calling scripts are calling multiple different scripts, some quite complex. I would want them to be standalone scripts and not be impacted by a large and completely different database context. And the subsequent scripts would be forced to use memory cached objects rather than reloading from the database. And for the sake of independence and avoiding unintended side effects, including performance, I would want each script to be standalone in its operation with the database.

As you know, when operating on a lot of records, the cached objects in a database context can get quite large and impact performance. By creating a new db context, you avoid this memory pressure.

As long as the connection string is identical, then the connection will be reused and the dbcontext list of pending changes to be made with the context will still be unique between scripts. So, there would be no connection resource hit.

Regarding the lack of HttpContext in Batch, the code uses the HttpRuntime.Cache with a unique SessionId (Guid) as index for each database in the absence of HttpContext. The new Context is created by the calling script using the database name (Host) from the calling script's active dbcontext (not from the HttpContext) so the connection string will be identical but have a separate dbcontext.

PR #401, opened 2019-02-12
Td goer funding page

timdavila:


CmsWeb/Areas/OnlineReg/Controllers/OnlineRegController.cs
```diff
@@ -31,7 +31,12 @@ public ActionResult Index(int? id, bool? testing, string email, bool? login, str
             {
                 var m = new OnlineRegModel(Request, CurrentDatabase, id, testing, email, login, source);
                 if (m.org != null && m.org.IsMissionTrip == true)
-                    m.PrepareMissionTrip(gsid, goerid);
+                {
+                    if (gsid != null || goerid != null)
```
josearmandohf: I think it is very picky but what about > if(gsid.IsNotNull() && goerid.IsNotNull())
Nothing that stops the PR from being approved but probably as a standard for refactoring processes.

PR #405, opened 2019-02-15
Td email special links

timdavila:
https://trello.com/c/lDv3IEko/5349-new-special-links-dialog-for-email-builder-feature


CmsWeb/Content/touchpoint/js/email/compose.js
```diff
@@ -266,6 +266,148 @@
...
+                linkText += '&text=' + insideText + '}}';
+            } else {
+                linkText = '';
+            }
+
+            specialLinks.resultInput.val(linkText);
+        }
+    };
+
+    $(document).ready(specialLinks.init);
```
brondavies: since you're already in a jQuery startup function you can just call `specialLinks.init();`

PR #408, opened 2019-02-20
Contribution Search should support using QueryTags as a filter

davidcarroll:
Contribution Search can filter by a normal Tag. This PR allows filtering by a QueryTag too.
QueryTag is a special type of tag that is only created via the PythonModel. It is used for caching a list of people for fast reporting and for easy validation and reconcilation. A QueryTag can be searched with SearchBuilder but it would be very helpful to filter contributions by the people in a QueryTag.
The only way to set a QueryTag as the active Tag is via a QueryString on the Contribution search URL called setQueryTag. This allows a link to be on the report that will go directly to the Contribution Search and run giving a total that should match up with the report.

### DIFF
--- a/CmsData/API/QueryFunctions/Query.cs
+++ b/CmsData/API/QueryFunctions/Query.cs
@@ -143,7 +143,12 @@ namespace CmsData
             var cn = GetReadonlyConnection();
             var parameters = new DynamicParameters();
             if (declarations != null)
+            {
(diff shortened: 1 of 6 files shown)


brondavies:
Closing because a previous PR already contains these changes and will be merged soon.

davidcarroll:
@brondavies\, this PR needs to stay as the previous PR does not have all this code.

PR #409, opened 2019-02-20
Refactor DbUtil.Db

davidcarroll:
`DbUtil.Db` was intended to be a pattern for a web application such that the `DbContext` is created during a Web Request and be kept available during the life of the Request. The instance of the `DbContext` is stored in `HttpContext`. This is a common and simple pattern to implement. 

The problem is that the implementation was located in `CmsData` library which invited misunderstanding and misuse outside of a WebApplication where there is no `HttpContext`.

There are times in an `Entity`, that you would like to have access to the `DbContext`. Unfortunately, there is no public reference to the current `DbContext` the `Entity` was created in. The obvious solution is to pass the `DbContext` in to method parameters whenever it is needed. That works and this PR has changed a number of methods to do just this. There are, however, a number of classes that have made extensive use of `DbUtil.Db` inside `CmsData`. Ostensibly, these cases were all intended to be used only by `CmsWeb`. If that is true, then they should have all been kept in `CmsWeb` and not put in `CmsData`. Regardless, I have replaced all such calls to `DbUtil.Db` with  `DbUtil.DbFromHttpContext` (a new method). This new method will only use `HttpContext` and not ever instantiate a new `DbContext`, but rather fail if it is used outside of a Web Request. This will enforce the proper approach during coding.

In the interest of avoiding all use of `DbUtil.Db`, I have created a new static class called `Ctx` in `CmsWeb` that should replace all use of `DbUtil.Db` in `CmsWeb` with a call to Ctx.Db.

So bottom line, there is no longer any code that uses `DbUtil.Db` neither in `CmsData` nor in `CmsWeb`. As such, it is ready to be deleted. I feel really good about this approach and it elevates `Ctx.Db` so that it can be trusted to behave properly since it will only be used in the intended way.

I have done basic testing by browsing around the application and exercising many operations and pages.


aternex:
👍

PR #411, opened 2019-02-22
Austin tag and untag modifications

aternex:
https://trello.com/c/TwxXSZmn/5350-2-issue-with-tag-all-in-search-builder-tags-all-people-records-becky-gossard-salemalliance-32046

### DIFF
--- a/CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
+++ b/CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
@@ -186,17 +186,10 @@ namespace CmsWeb.Areas.Search.Models
             Db.CurrentTagOwnerId = tag.PersonOwner.PeopleId;
 
             var q = DefineModelList();
-            var people = q.ToList();
 
-            foreach (var person in people)
-            {
-                if(person.Tags.Any(t => t.Id == tag.Id))
-                {
-                    continue;
-                }
-
-                Db.TagPeople.InsertOnSubmit(new TagPerson { PeopleId = person.PeopleId, Id = tag.Id, DateCreated = DateTime.Now });
-            }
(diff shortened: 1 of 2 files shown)


CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
```diff
@@ -152,57 +151,98 @@ public IEnumerable<CategoryClass> FieldCategories()
...
+            Db.CurrentTagName = tag.Name;
+            Db.CurrentTagOwnerId = tag.PersonOwner.PeopleId;
+
+            var q = DefineModelList();
+            var people = q.ToList();
+
+            foreach (var person in people)
             {
-                Db.TagAll(q, tag);
+                if(person.Tags.Any(t => t.Id == tag.Id))
```
brondavies: You should add this to the query instead of looping here because it will run another query for the `Tags` for each person
e.g. `q.Where(p => !p.Tags.Any(...))`

CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
```diff
@@ -152,57 +151,98 @@ public IEnumerable<CategoryClass> FieldCategories()
...
+            Db.CurrentTagName = tag.Name;
+            Db.CurrentTagOwnerId = tag.PersonOwner.PeopleId;
+
+            var q = DefineModelList();
+            var people = q.ToList();
+
+            foreach (var person in people)
             {
-                q = Db.PersonQueryParents(q);
+                if (person.Tags.Any(t => t.Id == tag.Id))
```
brondavies: same here e.g. `q.Where(p => p.Tags.Any(...))`

aternex: thank you, i missed these. correcting now!

PR #414, opened 2019-02-22
Td email special links

timdavila:


timdavila:
> So, no more `{{ }}`?

Right, not needed now that I can rely on the quotes in the anchor tag instead

PR #417, opened 2019-02-25
Multiple gateway DB structure

josearmandohf:
Multiple gateway DB structure


CmsData/Migrations/20190225_multiple_gw_db.sql
```diff
@@ -0,0 +1,180 @@
+IF EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES where
```
brondavies: The pattern we use is to create if not exist - particularly for tables - we don't want to lose data by accident.  Please switch this file to use the "if not exists" pattern.  Then your inserts to seed table data will also only happen the first time the table is created.

josearmandohf: @brondavies I have updated it accordingly.

PR #431, opened 2019-03-06
Fix the tagging and untagging for people from tagcontroller and query…

aternex:
Fix the tagging and untagging of people. I was incorrectly handling the creation of the TagPerson record and this was throwing an exception. Also, correct an issue when deleting from the QueryController instead of the TagController. QueryController uses the Util.CurrentTag and does not send the name of the tag. Now, the CurrentTag gets looked up instead.

### DIFF
--- a/CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
+++ b/CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
@@ -187,23 +187,28 @@ namespace CmsWeb.Areas.Search.Models
 
             var q = DefineModelList();
 
-            Db.TagPeople.InsertAllOnSubmit(q.Where(p => !p.Tags.Any(t => t.Id == tag.Id))
-                .Select(p => new TagPerson { PeopleId = p.PeopleId, Id = tag.Id, DateCreated = DateTime.Now })
-                .ToList());
+            var newTags = new List<TagPerson>();
 
+            foreach (var p in q.Where(p => !p.Tags.Any(t => t.Id == tag.Id)).ToList())
+            {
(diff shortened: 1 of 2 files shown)


CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
```diff
@@ -187,23 +187,28 @@ public Tag TagAll(Tag tag)
 
             var q = DefineModelList();
 
-            Db.TagPeople.InsertAllOnSubmit(q.Where(p => !p.Tags.Any(t => t.Id == tag.Id))
```
aternex: Exception was being thrown here on TagAll()... You cannot create the entity as part of the query directly. Adding a list to store the records and then add those records to the db.

CmsWeb/Areas/Search/Models/Query/Model/QueryModel.cs
```diff
@@ -187,23 +187,28 @@ public Tag TagAll(Tag tag)
...
         /// </summary>
-        /// <param name="tagname">The name of the tag to remove from the result of this QueryModel. Uses the user's session id as a default value if nothing is supplied</param>
+        /// <param name="tagname">The name of the tag to remove from the result of this QueryModel. Uses the current tag name as a default value if nothing is supplied</param>
         public Tag UntagAll(string tagname = "")
         {
             if (string.IsNullOrEmpty(tagname))
             {
-                tagname = Util.SessionId; // not specifying an explicit name, so use the session id as a default
+                tagname = Util2.CurrentTag; // not specifying an explicit name, so use the current tag name as default
             }
```
aternex: UntagAll, called from querycontroller, wasn't passing a tagname and it was using the sessionid. Based on the UI and the usage, it makes more sense to pass the currenttag.

PR #449, opened 2019-03-22
Fix two bugs

davidcarroll:
See the descriptions on these commits.
This addresses issues I posted a couple of days ago.


davidcarroll:
@brondavies 
It appears that the code in Util.Host and Util.GetConnection was ignored when creating this function.
There are a lot of special cases that Util is handling, particularly in this case the experience that a new GitHub user gets when starting up for the first time and when running locally in debug mode.
That is why I went back to using Util for this fix. If you don't want to rely on Util for things like this, please at least incorporate all the special cases that code handles, mainly DbServer and Host from AppSettings. I have spent many hours working around the issues with this one function for many days now, and now it is affecting GitHub users.

davidcarroll:
@brondavies I hope we can get this published soon. Mount Pisgah is not able to run one of their custom reports because of the timeout problem and GitHub users will fail on the first download and run. I see where you rewrote the Util functions and that looks fine to me, but all of these commits, besides that one that shows a conflict (the one you rewrote), need to be published. I hope you can do this soon.

davidcarroll:
I have a meeting Wednesday to go over everything with MP. It sure would be helpful to be able to demo GivingHouseholds script working in production. All that I need is the timeouts on the DataReaders in place. @brondavies

brondavies:
Closing because the changes were merged or implemented in other PRs

PR #462, opened 2019-03-29
Td anonymous supporters fix

timdavila:
https://trello.com/c/c3bYtzwE/5060-add-goer-funding-page-for-rush-creek


timdavila:
updated this a bit - please review the changes in the generated GoerSenderAmount class - I actually did those manually as I wasn't sure where those relations are defined in the db. The FK just points to the people table, and I couldn't find the "goer" and "sender" distinction. It looks like these were mixed up, looked elsewhere in the project, and it seems like it wasn't referenced before which is why this probably hasn't been spotted before

PR #474, opened 2019-04-11
Repair First Time Startup

davidcarroll:
These changes fix the experience of a GitHub user who downloads the code to match with the README
so that he just makes to two changes to web.config for dbserver and host and then clicks run.


davidcarroll:
Closing to resubmit

PR #475, opened 2019-04-12
Redeemer Community Group Fix

brianlankford1:
Moved error logic up the stack a bit and specified the model error key to be specific for the member selected to register.


brianlankford1:
@brondavies This is the fix for the redeemer community group error message.

PR #477, opened 2019-04-15
Td meeting types report

timdavila:
https://trello.com/c/RgVCK0H9/5479-meeting-types-report-for-fielder


CmsWeb/Views/Shared/Menu/ReportsMenu.xml
```diff
@@ -6,6 +6,7 @@
   <Header>Attendance Summary</Header>
   <Report link="/Reports/ChurchAttendance" target="_blank">Week at a Glance</Report>
   <Report link="/Reports/ChurchAttendance2" target="_blank">Average Week at a Glance</Report>
+  <Report link ="/Reports/MeetingAttendance" target="_blank">Attendance by Meeting Type</Report>
```
brondavies: I feel like this report should not be visible in the menu if the setting `AttendanceUseMeetingCategory` is `false`  What do you think?

timdavila: Yes, I left a comment on the trello card about this. I wasn't sure how to achieve this with the current xml setup for the navigation. It looks like currently it only supports limiting by role. Do I need to add a new param to support this or is there a simpler way?

brondavies: looks like adding a `setting` attribute to `CmsWeb\Code\ReportsMenu.cs` wouldn't be too hard.  Lines 134 - 148 show how roles work to exclude an item based on role.

brondavies: I guess the Trello card said to use the setting `UseMeetingDescriptionPickList` so let's stick with that

PR #479, opened 2019-04-16
this migration is giving problems when creating DB for provisioning s…

josearmandohf:
…ystem.


davidcarroll:
There is a fix for the migration problem in my PR

PR #481, opened 2019-04-16
removing bad migration

josearmandohf:


davidcarroll:
@aternex See the following for a fix to the migration problem. 
The regex to split the script between GO statements would get a false 'go' because of the following lines that start with 'go'. See the link to the diff at the bottom for the changed line that fixes the regex.
```
DECLARE @supportsToFix TABLE
(
ID INT IDENTITY(1,1),
orgDatum INT,
supporterDatum INT,
goerDatum INT,
amountDatum MONEY,
idGSA INT,
orgGSA INT,
supporterGSA INT,
goerGSA INT,
amountGSA MONEY
);
```
https://github.com/bvcms/bvcms/pull/474/commits/da55936858ab1b8d5778b78d77d15be7d1cf9d06#diff-529f8bc46f253c4beb47d257f497b514R281

PR #486, opened 2019-04-24
Add validations for mission trips.

dahnbaez:
Supporters can make a general donation even if they don't support a specific participant.

### DIFF
--- a/CmsWeb/Areas/OnlineReg/Models/OnlineRegPerson/ValidateModelQuestions.cs
+++ b/CmsWeb/Areas/OnlineReg/Models/OnlineRegPerson/ValidateModelQuestions.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


CmsWeb/Areas/OnlineReg/Models/OnlineRegPerson/ValidateModelQuestions.cs
```diff
@@ -80,6 +78,30 @@ public void ValidateModelQuestions(ModelStateDictionary modelstate, int i)
                 Log("QuestionsRetry");
         }
 
+        private void ValidateSupportGoerRequired(bool supportGoerRequired)
```
brondavies: Fix the complex method issue by splitting this up into two functions and move the `if (supportGoerRequired)` into `ValidateModelQuestions`
Like This:
```
if (supportGoerRequired) {
    ValidateSupportGoerRequired();
} else {
    ValidateSupportGoerNotRequired();
}
```

PR #488, opened 2019-04-25
Repair first time startup 2nd PR

davidcarroll:
I saw some conflicts in the first PR on this so I decided close and resubmit in a simpler form.
The conflicts evidently came after I had issued the PR.
Hope to get this one in before any other conflicts arise.
See Commit comment.


davidcarroll:
@brondavies I have a self-hosting church I am trying to help, not a touchpoint client, who has downloaded the code but is having problems with getting the application to run and it has to do with the issues this pull request addresses. I hope to get this into production soon.

davidcarroll:
@brondavies any idea when you will approve this?

brondavies:
Closing because we don't want to use the DbServer setting anymore.  If someone needs to change the connection string in web.config - it should be using the standard settings in connection strings.  Also disabling CLR security for the whole db server should be an intentional choice by the user (even though our code requires it currently)  We should add that to the setup documentation.

davidcarroll:
@brondavies 
The problem I was trying to solve again (since recent code changes broke this behavior) was that the typical person who downloads the source for their church is not a developer. I was wanting to help them "fall into the pit of success" by having a simple set up for first-time use. Expecting them to figure out connection strings can be a hindrance to quick success.

There is another reason I wanted to isolate `dbserver` and `host` for debugging purposes was to allow entering those values into `secrets.config` making it much more convenient for a developer (me in particular), to pull the latest source code without having to worry about overwriting my configuration.

I'm OK with rewriting the READ.ME file but go a bit further. What I would like to do is implement the scheme described in this StackOverflow answer: https://stackoverflow.com/a/3920462/5033
Will you approve that approach? Don't want to waste my time again, but I do want to make sure the process for a neophyte user is easy as it used to be (and make it better for the developer too).

PR #493, opened 2019-04-29
Td org name filter fix

timdavila:
https://trello.com/c/1SFu0okC/5546-bug-blank-space-prefixed-in-name-search-under-org-people-tab

### DIFF
--- a/CmsData/Organization/OrgFilter.cs
+++ b/CmsData/Organization/OrgFilter.cs
@@ -12,21 +12,8 @@ namespace CmsData
         public string NameFilter
         {
             get
-            {
-                string filter = "";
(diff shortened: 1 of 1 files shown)


CmsData/Organization/OrgFilter.cs
```diff
@@ -11,7 +11,23 @@ public partial class OrgFilter
     {
         public string NameFilter
         {
-            get { return $"{FirstName} {LastName}"; }
+            get
```
brondavies: What if we just did `return $"{FirstName} {LastName}".Trim();`?

timdavila: Yep, much simpler

PR #496, opened 2019-05-02
Add description column and map check number

aternex:


aternex:
Tested this locally and imported all rows successfully from the spreadsheet provided for testing after adding the missing Pledges worksheet.

PR #508, opened 2019-05-17
Daniel fix email confirmation prompt bug

dahnbaez:
https://trello.com/c/KIlZlCkN/5569-2-bug-email-double-sending-nathan-barley-bellevue-36000

### DIFF
--- a/CmsWeb/Content/touchpoint/js/email/Email.js
+++ b/CmsWeb/Content/touchpoint/js/email/Email.js
@@ -6,7 +6,24 @@
         $.block();
         $('#Body').text(CKEDITOR.instances["Body"].getData());
         var q = $(this).closest('form').serialize();
+        if ($(this).attr('data-prompt') === 'True') {
+            var count = $("#Count").val();
+            swal({
+                title: "Are you sure?",
+                text: "You're about to send an email to " + count + " people.",
+                type: "warning",
+                showCancelButton: true,
+                confirmButtonClass: "btn-confirm",
+                confirmButtonText: "Yes, send it!",
+                showLoaderOnConfirm: true,
+                closeOnConfirm: false
+            }, sendEmail(q));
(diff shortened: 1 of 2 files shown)


CmsWeb/Content/touchpoint/js/email/Email.js
```diff
@@ -6,7 +6,24 @@
...
+            swal({
+                title: "Are you sure?",
+                text: "You're about to send an email to " + count + " people.",
+                type: "warning",
+                showCancelButton: true,
+                confirmButtonClass: "btn-confirm",
+                confirmButtonText: "Yes, send it!",
+                showLoaderOnConfirm: true,
+                closeOnConfirm: false
+            }, sendEmail(q));
```
brondavies: this is supposed to be called as a closure function but instead it's just executing directly here.  You'll have to use `, function(){ sendEmail(q); });` instead.

CmsWeb/Content/touchpoint/js/email/Email.js
```diff
@@ -20,6 +37,7 @@
                 if (taskid === 0) {
                     $.unblock();
                     swal("Success!", ret.content, "success");
+                    $('button.Send').prop('disabled', true);
```
brondavies: better to scope this to the form as `$('button.Send', q)` so that it doesn't affect all `button.Send` elements in the whole page

CmsWeb/Content/touchpoint/js/email/Email.js
```diff
@@ -31,6 +49,7 @@
                                 if (ret.title == 'Email has completed.') {
                                     swal(ret.title, ret.message, "success");
                                     window.clearInterval(intervalid);
+                                    $('button.Send').prop('disabled', true);
```
brondavies: add scope here too.

PR #515, opened 2019-05-23
Organization Search Filter Fixed

jfescobar18:


CmsWeb/Areas/Org/Models/OrgSearchModel.cs
```diff
@@ -289,8 +289,18 @@ private List<OrgSearch> FetchOrgsList()
 
         public IQueryable<OrgSearch> FetchOrgs()
         {
-            var queryable = CurrentDatabase.OrgSearch(Name, ProgramId, DivisionId, TypeId, CampusId, ScheduleId, StatusId, OnlineReg,
+            IQueryable<OrgSearch> queryable = null;
+
+            if(CampusId == -1)
+            {
+                queryable = CurrentDatabase.OrgSearch(Name, ProgramId, DivisionId, TypeId, 0, ScheduleId, StatusId, OnlineReg,
```
brondavies: If you look at the `OrgSearch` sql function, it is written to handle a null `CampusId` so it would be simpler to just use a null if CampusId is -1

jfescobar18: I've tried that before but it didn't work, that's why I solved it just like this.

PR #521, opened 2019-05-28
Fran fix add people extra pk already exists

jfescobar18:


CmsData/Migrations/20190520_Fix_People_Extra_Constaint.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN TRAN FixConstraint
+	BEGIN TRY  
+		ALTER TABLE [PeopleExtra]
+		DROP CONSTRAINT [PK_PeopleExtra_1]
+
+		ALTER TABLE [PeopleExtra]
+		ADD  CONSTRAINT [PK_PeopleExtra_1] PRIMARY KEY CLUSTERED 
+		(
+			[PeopleId] ASC,
+			[TransactionTime] ASC,
```
brondavies: I don't think this is the right fix - most likely when the user enters information in a registration and updates a value targeting an extra value, it should update the EV instead of creating a new one

jfescobar18: Right, I'm working on that

PR #527, opened 2019-05-30
Tag All now tags only people from search result.

josearmandohf:


brondavies:
Really nice fix @josearmandohf 💯❗️

PR #544, opened 2019-06-11
Coupon errors fixed on paylink

jfescobar18:


CmsWeb/Areas/OnlineReg/Controllers/Coupon.cs
```diff
@@ -81,7 +81,7 @@ public ActionResult ApplyCoupon(PaymentForm pf)
             }
             var tid = $"Coupon({Util.fmtcoupon(coupon)})";
 
-            if (!pf.PayBalance)
+            if (c.Amount > pf.AmtToPay || !pf.PayBalance)
```
brondavies: should this be `>=` so that if the coupon applied is equal to the amount due it will not require payment?

jfescobar18: Yes you right

PR #560, opened 2019-06-19
make check display dismissable by keyboard or mouse

timdavila:
https://trello.com/c/M5H0hdeN/5662-2-check-display-locks-system-up-jeannie-lewis-mountpisgah-36639


brondavies:
This is a big improvement but I wasn't able to use the `ESC` key to dismiss the dialog.

timdavila:
that's strange, it works on my machine - are you sure you were loading the updated js?

brondavies:
Maybe the build didn't compile the scripts correctly.  I'll look into it

PR #567, opened 2019-06-26
fix date validation on mobile

timdavila:
https://trello.com/c/dGqjuyfX/5661-bug-contacts-made-created-on-phones-show-error-and-then-create-a-contact-with-missing-data


brondavies:
Got a unit test you can add?

timdavila:
Ah, ok I'll try to make one.

PR #569, opened 2019-06-27
add help to edit org settings, layout consistency

timdavila:
https://trello.com/c/oCo4kCB8/5677-3-suggestion-wayne-brown-ebcjackson-36743

### DIFF
--- a/IntegrationTests/Areas/Manage/AccountTests.cs
+++ b/IntegrationTests/Areas/Manage/AccountTests.cs
@@ -1,5 +1,7 @@
 ﻿using IntegrationTests.Support;
 using Shouldly;
+using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 
 namespace IntegrationTests.Areas.Manage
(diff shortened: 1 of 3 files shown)


IntegrationTests/Support/AccountTestBase.cs
```diff
@@ -62,6 +62,7 @@ protected CmsData.User CreateUser(CmsData.Family family = null, params string[]
             if (roles.Any())
             {
                 user.AddRoles(db, roles);
+                db.SubmitChanges();
```
brondavies: nice catch 👍

IntegrationTests/Areas/Manage/AccountTests.cs
```diff
@@ -73,5 +75,33 @@ public void MyData_User_ChangePassword_Test()
             PageSource.ShouldContain(user.Person.Name);
             PageSource.ShouldContain(user.Person.EmailAddress);
         }
+
+        [Fact]
+        public void Create_Role_Test()
```
brondavies: We might move this to a new test class `RoleTests` or something later but it's fine here for now.

PR #571, opened 2019-07-01
Fix Acceptiva batch issues

dahnbaez:
Removed "Touchpoint" word in the transaction id Acceptiva was returning.

### DIFF
--- a/CmsData/Finance/AcceptivaGateway.cs
+++ b/CmsData/Finance/AcceptivaGateway.cs
@@ -258,7 +258,7 @@ namespace CmsData.Finance
             {
                 batchTransactions.Add(new BatchTransaction
                 {
-                    TransactionId = int.Parse(item.Response.Items[0].Id),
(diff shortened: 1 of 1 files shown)


CmsData/Finance/AcceptivaGateway.cs
```diff
@@ -258,7 +258,7 @@ public BatchResponse GetBatchDetails(DateTime start, DateTime end)
             {
                 batchTransactions.Add(new BatchTransaction
                 {
-                    TransactionId = int.Parse(item.Response.Items[0].Id),
+                    TransactionId = int.Parse(item.Response.Items[0].Id.Remove(0, 11)),
```
brondavies: Can you use `Replace()` with a specific string so we know what is getting removed. Then, if for some reason they stop adding that to the transaction ID, it won't break.

dahnbaez:
This fix is now at pr575

PR #590, opened 2019-07-25
fix for empty url

timdavila:
https://trello.com/c/PewhWaZ4/5729-2-funding-page-issue-maddie-huang-fielder-37097

Still not sure what happened here. I tried to add a test for this, but couldn't figure out how to mock the HttpRequest in unit tests as if it were coming through OnlineRegController without the RequestBase


brondavies:
> Still not sure what happened here. I tried to add a test for this, but couldn't figure out how to mock the HttpRequest in unit tests as if it were coming through OnlineRegController without the RequestBase

You would do it similar to what's implemented here: https://github.com/bvcms/bvcms/blob/develop/UnitTests/UtilityExtensionsTests/UtilSessionTests.cs#L16

PR #595, opened 2019-08-02
Temporary Payment Gateway not Auto Configured

jfescobar18:


CmsWeb/Areas/Setup/Controllers/GatewayController.cs
```diff
@@ -140,9 +140,9 @@ public JsonResult InsertAccount([System.Web.Http.FromBody]Models.GatewayAccountJ
 
                 if (json.UseForAll)
                 {
-                    for (int i = 0; i < CurrentDatabase.PaymentProcess.Select(x => x.ProcessId).Count(); i++)
+                    for (int i = 1; i < CurrentDatabase.PaymentProcess.Select(x => x.ProcessId).Count(); i++)
                     {
-                        paymentProcess = CurrentDatabase.PaymentProcess.SingleOrDefault(x => x.ProcessId == i + 1);
+                        paymentProcess = CurrentDatabase.PaymentProcess.SingleOrDefault(x => x.ProcessId == i);
```
brondavies: This doesn't solve the problem.  You also missed the other case for adding a new payment gateway https://github.com/bvcms/bvcms/pull/595/files#diff-74b4603ea7cbeb0ea947e18ad34497e2R116

brondavies: Never mind... I see it now - the code just didn't make sense

PR #598, opened 2019-08-05
Columbarium project

davidcarroll:
I'll post the scripts and such to the documentation repo after those have settled in.


davidcarroll:
@brondavies I have fixed the Release Build.
Also, I published the Python scripts and other files to https://github.com/davidcarroll/Columbarium

CmsWeb/Controllers/ScriptController.cs
```diff
@@ -152,179 +95,131 @@ public ActionResult RunScriptExcel(string scriptname, string parameter = null)
             }
         }
 
-        [HttpGet, Route("~/PyScript/{name}")]
+        [HttpGet, Route("~/PyScript/{name}/{p1?}/{p2?}")]
```
brondavies: Does changing this route mean that existing /PyScript URLs that have p1= and p2= in the query string will fail?  Can we stay compatible with the existing behavior?

davidcarroll:
It should not be a problem since they are marked as optional with the "p1?"
and "p2?"

davidcarroll:
https://stackoverflow.com/a/31257183/5033

PR #600, opened 2019-08-07
Fran family id added to involvement

jfescobar18:


UnitTests/CMSWebTests/Areas/Reports/Models/Reports/Export/ExportInvolvementsTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        public void InvolvementList_Should_Have_FamilyId()
+        {
+            var db = CMSDataContext.Create(Util.Host);
+            var m = OrganizationModel.Create(db, FakeRequestManager.FakeRequest().CurrentUser);
+            var FakeOrg = FakeOrganizationUtils.MakeFakeOrganization();
+            
+            m.OrgId = FakeOrg.org.OrganizationId;
+
+            var TestInvolvementList = ExportInvolvements.InvolvementList(m.QueryId);
+            TestInvolvementList.ShouldNotBeNull();
```
brondavies: Here you should test that the data in `TestInvolvementList` also contains the expected values (not just checking for null)

UnitTests/CMSWebTests/Areas/Reports/Models/Reports/Export/ExportInvolvementsTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        [Fact]
+        public void ShouldDeleteReg()
+        {
+            FakeOrganizationUtils.DeleteOrg(OrgId);
```
brondavies: Instead of creating a separate test, make this part of the `InvolvementList_Should_Have_FamilyId` test because they might not run in the order you expect.

PR #610, opened 2019-08-22
no dob and no phone controlled by master org

jfescobar18:


CmsWeb/Areas/OnlineReg/Models/OnlineRegPerson/Helper.cs
```diff
@@ -737,6 +753,12 @@ public void Log(string action)
             DbUtil.LogActivity("OnlineReg " + action, masterorgid ?? orgid, PeopleId ?? Parent.UserPeopleId, Parent.DatumId);
         }
 
+        public bool IsInMasterOrg()
+        {
+            var o = masterorg ?? org;
+            return db.Organizations.Select(x => x.OrgPickList).Contains(o.OrganizationId.ToString());
```
brondavies: You can't rely on this because `"7,8,9".Contains("7")` is true and so does  `"17,18,19".Contains("7")` 
Instead use `Organizations.Where(o => db.SplitInts(o.OrgPickList).Any(i => i.ValueX == o.OrganizationId)`

PR #621, opened 2019-08-27
SearchBuilder option to show all orgs without specifying Program and Division

davidcarroll:
There is a new Setting called ShowAllOrgsByDefaultInSearchBuilder will default to true.
Set to false if old behavior is desired.


davidcarroll:
By the way, I think that @msmith-tp will want to blog about this. I think this will really be a convenience for people who work with SearchBuilder a lot.

msmith-tp:
@davidcarroll Am I understanding correctly that, with this new setting, any Search Builder condition that asks for program / division will default to all organizations if prog/div are left blank?

davidcarroll:
Correct, and you can immediately do a search for the one org you want
without fooling with Program and Division. I could never remember which
Prog/Div the org was in and even if I could, it was a hassle to do so many
clicks and scrolls through the dropdowns just to get to the one org. Should
be helpful for any church I think. You can still set a Program and/or
Division to find members in an entire Program or Division. But when you are
just wanting a single org, it will be super easy.

davidcarroll:
The only reason I made it an option was in case there were any repercussions for extremely large churches, you could turn it off. Bellevue has 6000+ orgs, and that may be problematic for the page size, but I don't think it will be a problem.

davidcarroll:
@msmith-tp I just wanted to make sure you saw the two previous comments in reply to yours in case you weren't subscribed to the notifications since I forgot to reference your name.

PR #622, opened 2019-08-27
Assign CurrentDatabase during ModelBinding

davidcarroll:
This will avoid having to do: `m.Db = CurrentDatabase;`
in the controller for models that are bound during POST.
Included is a UnitTest to verify that this is working correctly.


davidcarroll:
@brondavies Glad I can still contribute!

PR #651, opened 2019-09-16
picture directory update

timdavila:
https://trello.com/c/nmzm2nnz/5711-bug-directory-includes-deceased-children-2-photo-directory-westendcrc-37008

### DIFF
--- a/UnitTests/CMSWebTests/CMSWebTests.csproj
+++ b/UnitTests/CMSWebTests/CMSWebTests.csproj
@@ -160,10 +160,6 @@
       <Project>{512da381-e81b-4121-b978-fc850ec88353}</Project>
       <Name>ImageData</Name>
     </ProjectReference>
-    <ProjectReference Include="..\..\IntegrationTests\IntegrationTests.csproj">
-      <Project>{298AF160-14F1-4CA2-8413-F74E100DF3EB}</Project>
-      <Name>IntegrationTests</Name>
-    </ProjectReference>
     <ProjectReference Include="..\..\UtilityExtensions\UtilityExtensions.csproj">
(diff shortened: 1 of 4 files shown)


UnitTests/CMSWebTests/TestUtils/FakeOrganizationUtils.cs
```diff
@@ -73,5 +74,17 @@ public static void DeleteOrg(int OrgId)
                     .PurgeOrg(db);
             }
         }
+
+        public static string RandomString(int length = 8, string prefix = "")
```
brondavies: Please use the implementation in `DatabaseTestBase` instead of making a copy.  You can make it public.

UnitTests/CMSWebTests/CMSWebTests.csproj
```diff
@@ -160,6 +160,10 @@
       <Project>{512da381-e81b-4121-b978-fc850ec88353}</Project>
       <Name>ImageData</Name>
     </ProjectReference>
+    <ProjectReference Include="..\..\IntegrationTests\IntegrationTests.csproj">
+      <Project>{298AF160-14F1-4CA2-8413-F74E100DF3EB}</Project>
+      <Name>IntegrationTests</Name>
+    </ProjectReference>
```
brondavies: All test libraries should be independent from each other - shared test classes belong in `SharedTestFixtures`

brondavies: I've changed the test added to depend on `DatabaseTestBase` instead of `FeatureTestBase`.  The latter is only for tests run with Selenium automation.

PR #652, opened 2019-09-17
Fix Acceptiva testing tag issue

dahnbaez:
Verify if a Transaction it's an Acceptiva testing transaction in order to pull batches correctly

### DIFF
--- a/CmsWeb/Areas/OnlineReg/Models/OnlineReg/PaymentForm.cs
+++ b/CmsWeb/Areas/OnlineReg/Models/OnlineReg/PaymentForm.cs
@@ -711,6 +711,8 @@ namespace CmsWeb.Areas.OnlineReg.Models
 
             ti.TransactionId = tinfo.TransactionId;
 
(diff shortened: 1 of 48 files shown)


CmsWeb/Areas/OnlineReg/Models/OnlineReg/PaymentForm.cs
```diff
@@ -736,6 +738,16 @@ public Transaction ProcessPaymentTransaction(OnlineRegModel m)
             return ti;
         }
 
+        private bool? CheckIfIsAcceptivaTesting(bool testing, PaymentProcessTypes processType)
+        {
+            var gatewayId = MultipleGatewayUtils.GatewayId(CurrentDatabase, processType);
+            if (gatewayId == (int)GatewayTypes.Acceptiva && MultipleGatewayUtils.GatewayTesting(CurrentDatabase, ProcessType))
```
brondavies: @dahnbaez  Couldn't you just check whatever gateway testing value is here without specifying Acceptiva?  Then it would work the same for all gateways.

PR #657, opened 2019-09-24
Migration added and issue fixed

jfescobar18:


CmsData/Migrations/20190924_Update_Email_RecReg.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE r
+SET r.[email] = p.[EmailAddress]
+FROM [RecReg] as r
+RIGHT JOIN [People] as p
+ON r.[PeopleId] = p.[PeopleId]
+WHERE r.[PeopleId] = p.[PeopleId]
```
brondavies: @jfescobar18 will you also include `AND r.[email] IS NULL` so that we don't overwrite anything that shouldn't be changed?

PR #660, opened 2019-09-26
DOB and Phone on Find showing by default

jfescobar18:


jfescobar18:
> @jfescobar18 What about the existing organization settings?

I'll work on a migration for that

PR #667, opened 2019-10-01
Setting added to hide Deceased

jfescobar18:


CmsWeb/Areas/People/Views/Person/Family/Members.cshtml
```diff
@@ -41,21 +41,25 @@
...
-                                {
-                                    <text>&bull; <span class="status">@m.MemberStatus</span></text>
-                                }
-                                &bull; <span class="role">@m.PositionInFamily</span>
-                            </span>
-                            <div class="email email_display truncate">@m.Email</div>
-                        </a>
-                    </li>
+
+                    if (Model.Person.ShowDeceasedMember() && m.IsDeceased == false)
```
brondavies: should be `||` not `&&`

CmsData/Extensions/Person.cs
```diff
@@ -2487,5 +2487,11 @@ private static bool IsHeadOfHouseholdFromMyDataPage
                 return fam?.IsHeadOfHouseold(Util.UserPeopleId) == true;
             }
         }
+
+        public bool ShowDeceasedMember()
+        {
+            bool res = DbUtil.Db.Setting("HideDeceasedFromFamily", false) ? true : false;
+            return res;
+        }
```
brondavies: `public bool ShowDeceasedMember() => DbUtil.Db.Setting("HideDeceasedFromFamily");`
but it would be even better to find a way NOT to use `DbUtil.Db` perhaps by setting this as a `ViewBag` property for the view in the controller.

PR #669, opened 2019-10-02
allow deletion of contribution with tags

timdavila:
https://trello.com/c/QIQUmujc/5839-bug-duplicate-contribution-created-in-bundle-when-changing-bundle-type-and-then-closing-the-bundle


CmsWeb/Areas/Finance/Models/PostBundleModel.cs
```diff
@@ -21,13 +21,17 @@ public class PostBundleModel
     {
         private BundleHeader bundle;
 
+        private CMSDataContext CurrentDatabase;
+
         public PostBundleModel()
         {
+            this.CurrentDatabase = DbUtil.Db;
```
brondavies: Implement `IDbBinder` and you shouldn't have to do this.  Just make sure none of the calling code uses this constructor

timdavila: I need the empty constructor for the controller methods using DI. Unless there's another way?

brondavies: Implement `IDbBinder` @timdavila

PR #684, opened 2019-10-21
handle duplicate transactions on registrations

timdavila:
https://trello.com/c/iiJLGPlF/5920-1-credit-cards-being-declined-when-registering-two-children-for-similar-events-laurie-ennis-ingleside-38639


brondavies:
Needs a unit test @timdavila Thanks.

PR #686, opened 2019-10-22
fix capitalization as a first step to build on Linux

tpokorra:
We are a group of Linux enthusiasts in Germany who are members of various churches (https://luki.org/).
We are currently investigating bvcms, to see if it would work for German churches.
During this investigation, we are trying to build and run bvcms on Linux with Mono.
This Pull Request is just a small step, but should not affect the build on Windows at all.
We just replaced System.configuration with System.Configuration in some csproj files, and fixed paths and filename in resource files, and fixed the path Content in the gulpfile.

More work is in my fork at: https://github.com/SolidCharity/bvcms/tree/TP-20191017-build-fedora30
But that is not ready for a pull request at all. Just to let you know...

### DIFF
--- a/CmsCheckin/CmsCheckin.csproj
+++ b/CmsCheckin/CmsCheckin.csproj
@@ -162,7 +162,7 @@
       <HintPath>..\packages\Newtonsoft.Json.12.0.2\lib\net40\Newtonsoft.Json.dll</HintPath>
     </Reference>
     <Reference Include="System" />
-    <Reference Include="System.configuration" />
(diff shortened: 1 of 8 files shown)


brondavies:
Thanks @tpokorra We look forward to more of your contributions.  I'm going to approve this PR, but in the future if you have changes in the code, we'll be looking for unit tests.  When you have a working build, we'll incorporate Linux into our supported platforms and make it part of the daily build process to catch any breaking changes on Linux.  Thank you again.

PR #687, opened 2019-10-22
Address format changed

jfescobar18:


CmsWeb/Areas/Reports/Models/Avery/AveryAddressResult.cs
```diff
@@ -121,8 +121,8 @@ public override void ExecuteResult(ControllerContext context)
                     ph.AddLine(m.MailingAddress.Trim(), font); 
                 else
                 {
-                    ph.AddLine(m.Address, font);
                     ph.AddLine(m.Address2, font);
+                    ph.AddLine(m.Address, font);
```
brondavies: This isn't the right fix - I think it has more to do with the way our address correction web service handles the case we're trying to fix.

PR #703, opened 2019-11-01
update regsetting option class

timdavila:
https://trello.com/c/pq5JRnLA/5909-bug-getting-error-when-running-registration-settings-options-report


CmsData/Generated/Views/RegsettingOption.cs
```diff
@@ -277,22 +275,8 @@ public string OrganizationName
 
 		}
 
-        [Column(Name = "AskPassport", Storage = "_AskPassport", DbType = "bit")]
```
brondavies: What happened to AskPassport?

PR #710, opened 2019-11-12
Vital Stats Graphs Report FIX

josearmandohf:
Updating query to filter out Returned/Reversed contributions.


brondavies:
@josearmandohf Tests are failing with `System.NotSupportedException : An IQueryable that returns a self-referencing Constant expression is not supported.`  You may need to rework the queries that you changed.

PR #714, opened 2019-11-12
Admins can't see the record of a deceased person in their own family

jfescobar18:
https://trello.com/c/kNBHz1Ve


timdavila:
I'm not sure if this was already discussed, but this seems like it could be a a good use case for the role manager rather than a global setting and conditional role based logic. This way it could be more granular. Im also wondering if the default should be to hide them rather than show.

PR #724, opened 2019-11-26
Profile>Involvement>Registrations edit btn not respecting position.

josearmandohf:
Profile>Involvement>Registrations: placing html parent element outside of "@if(goerlist.Count > 0)" to respect element positioning.


josearmandohf:
The Edit Button should be at the top of the page, not between the Event Registration Fulfillment and Mission Trip Funding. Looks bad and is an awkward user experience. 

Developer Info.
Github branch: ja-ui-editbtn-display-fix
Github PR: https://github.com/bvcms/bvcms/pull/724
Test Site: pr724.tpsdb.work
Trello: https://trello.com/c/7kwclZb3/6015-bug-involvementregistrations-tabedit-button-appears-on-wrong-spot

PR #751, opened 2020-01-07
Fran email templates validate name so that empty template can not be used

jfescobar18:
https://trello.com/c/sVWvFImR


CmsWeb/Areas/Manage/Views/Display/Index.cshtml
```diff
@@ -397,6 +404,7 @@
 
             $('#new-modal-role').on('shown.bs.modal', function () {
                 $("#newName").val('').focus();
+                $("#newName-role").val('').focus();
```
brondavies: don't focus this field if you are focusing `#newName`

timdavila: You can also now autofocus an input natively in HTML:
https://www.w3schools.com/tags/att_input_autofocus.asp

PR #762, opened 2020-01-23
Html statement layouts

brondavies:


CmsWeb/Properties/Files/ContributionStatementTemplate.html
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <style type="text/css">
+        * {
+            font-family: Helvetica, Helv;
```
timdavila: Since we're not pulling in web fonts, I would include some other fallbacks, even if it's just `sans-serif`

brondavies: This renders with WKHTML on the server, so we get to control the availability of certain fonts on the system.  The old statement code used Helvetica as the font for statements too.

CmsData/Generated/Functions/NormalContribution.cs
```diff
@@ -24,11 +30,43 @@ public partial class NormalContribution
...
+            ContributionType = c.ContributionType;
+            BundleType = c.BundleType;
+            FundName = c.FundName;
+            CheckNo = c.CheckNo;
+            Name = c.Name;
+            Description = c.Description;
+            FundDescription = c.FundDescription;
+        }
+
+        public static explicit operator NormalContribution(NonTaxContribution c) => new NormalContribution(c);
```
josearmandohf: NonTaxContribution is being casted explicity as NormalContribution(c), just wondering if GiftsInKind needs to explicity be casted as => new NormalContribution(c) also.

brondavies: No, because those are treated differently, I didn't need to

PR #765, opened 2020-01-28
HttpRuntime.Cache.SessionId no expiration

jfescobar18:
https://touchpointsoftware.atlassian.net/browse/TPD-339?atlOrigin=eyJpIjoiMDVmZWU1Mzc1OWNhNDRmNjg4MGY3MzcxYjYxMDRlZjgiLCJwIjoiaiJ9


UtilityExtensions/Extensions/UtilityExtensions.cs
```diff
@@ -40,7 +41,7 @@ public static string SessionId
             {
                 return HttpContextFactory.Current?.Session?.SessionID ?? (string)HttpRuntime.Cache["SessionId"] ?? (SessionId = Guid.NewGuid().ToString());                
             }
-            set { HttpRuntime.Cache["SessionId"] = value; }
+            set { HttpRuntime.Cache.Insert("SessionId", value, null, DateTime.Now.Add(new TimeSpan(6, 0, 0)), Cache.NoSlidingExpiration); }
```
brondavies: This setter isn't actually used anywhere @jfescobar18 This can't be the fix you're looking for.  Also, `HttpRuntime.Cache` is an application-wide cache and shouldn't be used at all for user-specific data.

brondavies:
Not needed

PR #769, opened 2020-01-30
Acceptiva Batch Dates Refactor

dahnbaez:
This one is to test the Acceptiva API response in live environment.


brondavies:
@dahnbaez Do we need this code anymore or can we close this PR without merging?

PR #783, opened 2020-02-10
TPD-378 New way to add person

asherlloyd:
Enhances the existing 'Add Person' feature to include a fuzzy/basic search as well as an advanced search, as per TPD-378 requirements.


UtilityExtensions/Extensions/StringExtensions.cs
```diff
@@ -210,9 +210,22 @@ public static string RandomPassword(int length)
...
         {
-            if ((name ?? "").Contains(","))
+            if ((name ?? "").Contains(" ") && (name ?? "").ToCharArray()[0] != ' ') // search first name only
             {
-                var a = (name ?? "").Split(',');
+                var a = (name ?? "").Split(' ');
+                first = "";
+                if (a.Length > 1)
+                {
+                    first = a[0].Trim();
```
josearmandohf: This code is not removing the comma for comma sepparated names when assigning to first and last variables.
i.e. 
"Carroll, David" is set as first= "Carroll," and last = "David"
"David, Carroll" is set as first= "David," and last = "Carroll"

asherlloyd: I think this is addressed in #785

CmsWeb/Content/touchpoint/js/form-ajax.js
```diff
@@ -161,7 +161,7 @@
         return false;
     });
 
-    $('body').on('click', 'form.ajax a.ajax', function (event) {
+    $('body').on('click', 'form.ajax a.ajax, .search-add-person', function (event) {
```
brondavies: Could you just add the `ajax` class to the `a` element here: https://github.com/bvcms/bvcms/pull/783/files#diff-d883acede567a54b8dc9eea3181fb816R30

PR #790, opened 2020-02-14
TPD-253 Adds AutoSyncBatchDatesWindow setting to automatic batch check

asherlloyd:
https://touchpointsoftware.atlassian.net/browse/TPD-253

### DIFF
--- a/CmsData/DbUtil/Context.cs
+++ b/CmsData/DbUtil/Context.cs
@@ -1979,7 +1979,7 @@ This search uses multiple steps which cannot be duplicated in a single query.
             };
 
             var today = DateTime.Now;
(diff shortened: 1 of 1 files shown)


CmsData/DbUtil/Context.cs
```diff
@@ -1979,7 +1979,7 @@ public void UpdateStatusFlags()
             };
 
             var today = DateTime.Now;
-            var yesterday = today.AddDays(-1);
+            var range = today.AddDays(-DbUtil.Db.Setting("AutoSyncBatchDatesWindow").ToInt());
```
brondavies: The build failed because `yesterday` is still referenced below.

PR #826, opened 2020-03-17
Update README.md

jason-rice:
Added instructions to help if the user doesn't have Gulp installed. 
Also, added some instructions with examples for setting up the connection info to SQL database and server.


README.md
```diff
@@ -48,12 +49,16 @@ BVCMS is maintained by TouchPoint Software, Inc.
 
 1. Find and edit the web.config file in the **CMSWeb** folder.
 
+    - In the `appSettings` section, configure the `dbserver` value to point to the server that your database will be hosted on. This value will become part of your connection string.
```
brondavies: We don't actually use this setting anymore - it is deprecated.  It is recommended to update the CMS connection string

README.md
```diff
@@ -48,12 +49,16 @@ BVCMS is maintained by TouchPoint Software, Inc.
...
+    - For example, you could change \<add key="dbserver" value="" /> to \<add key="dbserver" value="localhost\SQLEXPRESS" />.
     - In the `appSettings` section, configure the `host` value to point to the name you would like to call your database (without the CMS_). This value will become part of your connection string.
+    - For example, you could change \<add key="host" value="" /> to \<add key="host" value="myBvcms" />.
     - In the `mailSettings` section, configure the pickupDirectoryLocation to a directory on your development machine (for testing purposes).
     - If you want to create a database with 150 test records in it, just use ``testdb`` for the host value.
 
 1. Find and edit the ConnectionStrings.config file in the **CMSWeb** folder
     - Change the database server in the connection string named `CMS` if `(local)` or `localhost` is not the name of your SQL Server. For instance, if you installed SQL Express with the default options, you would change `(local)` to `.\SQLEXPRESS`
+    - For example, \<add name="CMS" connectionString="Data Source=localhost\SQLEXPRESS;Initial Catalog=myBvcms;Integrated Security=True" />
```
brondavies: Our database names are always `CMS_{host}`.  You might use `Initial Catalog=CMS_myBvcms` and point that out here as well but the database name will always be changed in the connection string to match the `CMS_{host}` pattern even if you specify a database name here.

README.md
jason-rice: I've removed this sentence.

README.md
jason-rice: I removed this sentence as well.

PR #827, opened 2020-03-17
TPD-692 Add QuerySqlJsonArray Support for multi-line chart

davidcarroll:
touchpointsoftware.atlassian.net/browse/TPD-692

### DIFF
--- a/CmsData/API/QueryFunctions/Query.cs
+++ b/CmsData/API/QueryFunctions/Query.cs
@@ -1,13 +1,14 @@
-using System;
-using System.Collections.Generic;
-using System.Configuration;
-using System.Data.Common;
-using System.Data.SqlClient;
-using System.Linq;
-using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


CmsData/API/QueryFunctions/Query.cs
```diff
@@ -278,6 +279,28 @@ public string SqlNameCountArray(string title, string sql)
 ]";
         }
 
+        public string QuerySqlJsonArray(string sql)
+        {
+            using (var rd = db.Connection.ExecuteReader(sql))
```
brondavies: Use `db.ReadonlyConnection()` here instead.

davidcarroll: Good catch @brondavies

davidcarroll:
@brondavies I am not seeing the QuerySqlJsonArray function in Develop branch. The project for the AttendanceTrends multiline chart project depends on this.

PR #832, opened 2020-03-20
TPD-641 Multiple gateway refund fix

dahnbaez:
Return gateway for registration when organization description is not found
https://touchpointsoftware.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=TPD-641&assignee=5e19301c85a8c90ecaca7d44


CmsData/Finance/MultipleGatewayUtils.cs
```diff
@@ -103,9 +103,9 @@ public static PaymentProcessTypes ProcessByTransactionDescription(CMSDataContext
 
             switch (RegistrationTypeId)
             {
-                case null:
                 case 8:
                     return PaymentProcessTypes.OneTimeGiving;
+                case null:
```
josearmandohf: I don't think this change is going to do anything. You are just moving the case null two lines below from place but since the switch is going to evaluate only one case per time disregarding the position.

dahnbaez: The problem was that sometimes the transaction description doesn't match with any organization (due to a name change or org itself was removed) so registration type is null. the previous version of the code was taking the OneTimeGiving process when registration type is null. Instead of that, this little change will return the OnlineRegistration process when the registration type is null. I added a couple of unit tests to check that we are getting the correct process.

PR #833, opened 2020-03-23
TPD-447 Org filter update

asherlloyd:
Fixes filtering issues


CmsData/Organization/OrgFilter.cs
```diff
@@ -27,8 +27,8 @@ public string NameFilter
                 }
                 else
                 {
-                    FirstName = value; // treat FirstName as a fuzzy search
-                    LastName = null;
+                    FirstName = null;
+                    LastName = value; // treat FirstName as a fuzzy search
```
brondavies: Looks good, just need to change this to say "treat LastName as a fuzzy search"

PR #839, opened 2020-03-26
TPD-754 Donor Totals By Range report Fix 2

josearmandohf:
Updating ContributionSearch() function calls.


CmsData/Migrations/20200326_alter-GetTotalContributionsAgeRange-func.sql
```diff
@@ -0,0 +1,47 @@
+USE [CMS_concordjc]
```
brondavies: Need to remove

PR #840, opened 2020-03-27
TPD-441 Pushpay Anonymous User Fix

dahnbaez:
include excpetion in authenticity for pushpay methods


CmsData/Finance/MultipleGatewayUtils.cs
```diff
@@ -87,7 +87,11 @@ select new
         public static bool GatewayTesting(CMSDataContext db, PaymentProcessTypes processType)
         {
             var User = db.Users.SingleOrDefault(us => us.UserId == db.UserId);
+#if DEBUG
```
josearmandohf: Please remove the #if DEBUG block

PR #844, opened 2020-03-27
TPD-611 Vital Stats Graphs Error

jason-rice:
fixed bug by excluding contribution records with no contribution date. Contribution date in the Contribution table is not null able in the db. Theoretically, therefore, should be no records with  null contribution date. There is one record in the training db in the contribution table with a null in the contribution date.


CmsData/Extensions/Contribution.cs
```diff
@@ -129,7 +129,7 @@ public static List<Contribution> GetContributions(CMSDataContext db, int[] fundi
```
brondavies: @jason-rice  I would add the not null check here instead of below so that it filters out the bad records with or without fund ids selected.

jason-rice: Sure, not a problem. Great suggestion. Not sure why I chose to filter the results instead of the query.

PR #852, opened 2020-04-03
TPD-773 - Quick-search-not-working-on-mobile-screen-size

jason-rice:
fixed this problem by replicating the search box on desktop view for the mobile view. Then adding a new and separate vue instance for the mobile view search box, just like the vue instance for the desktop view search box. This now works correctly on my local machine!


CmsWeb/Content/touchpoint/js/vue-search-add.js
```diff
@@ -149,3 +149,7 @@ Vue.component("search-add", {
 new Vue({
     el: "#search-add-component"
 })
+
+new Vue({
```
timdavila, line 153: You can also initialize the Vue instance on a parent element of both components, and have both components live in the same instance. This will save some memory and allow sharing of data

jason-rice: Great point! I'll go ahead and make the change. Thanks!

PR #855, opened 2020-04-06
TPD-804 Fix for Saved Draft not saving Subject

davidcarroll:
Fixes when saving draft and when selecting draft


UnitTests/CmsDataTests/Properties/Text/RenderTemplateTest.txt
```diff
@@ -1,6 +1,6 @@
 <p>{{header.DateFrom}} - {{header.DateTo}}</p>
 {{#each results}}
-    <p>{{trans_date}}</p>
+    <p>{{Fmt trans_date "MM/dd/yyyy hh:mm:ss tt"}}</p>
```
davidcarroll, line 3: @brondavies This was one of the tests that was failing for me. The years were getting shortened to yy instead of yyyy so I just put an explicit format so that the test should pass regardless of culture or other options that may be set on the operating system.

PR #862, opened 2020-04-07
TPD-306 checkin conditional label entries org based on extra value and person status flags

timdavila:
https://touchpointsoftware.atlassian.net/browse/TPD-306


CmsWeb/Areas/Public/Models/CheckInAPIv2/LabelEntry.cs
```diff
@@ -32,8 +33,39 @@ public class LabelEntry
 
 		public LabelEntry( AttendanceCacheSet cacheSet, LabelFormatEntry formatEntry, Attendance attendance, AttendanceGroup group = null, int index = 0 )
 		{
+            bool conditionalRemovedEntry = false;
 			typeID = formatEntry.typeID;
 
+            if (formatEntry.orgEV != null) {
```
brondavies, line 39: Should probably be `formatEntry.orgEV.HasValue()` to account for empty string

PR #865, opened 2020-04-09
Joseph Robinson Evaluation - Handle Saved Searches Missing Condition

fmjnax:
**TECH TEST**

Saved Searches with Conditions that do not exist will throw hard fault errors or display incomplete views. Added text to the incomplete views to indicate the issue and perform dictionary checks to halt the hard fault exception and allow flow to continue. 

*NOTE* - After spending a few hours running through the application code, hooking it up to Application Insights, and monitoring as I navigated, I did not find much of anything substantial without having business context (which is a breath of fresh air for what I am used to!). I stumbled upon the Saved Searches issue, but I assume it is an artifact of the testdb setup. Initially this looked like a quick fix bug. However, it ended up being a bigger can of worms the deeper I dove. As a result, I do **NOT** suggest  this as a permanent solution. However, given the circumstances for this exercise it is at least a viable mitigation.


brondavies:
Thanks Joseph.  Nice work!

PR #873, opened 2020-04-20
TPD-789 Transaction Batches not displaying correctly

jason-rice:
All that was needed was to embed the flag in the razor page itself, so it could persist from one server call to the next. Looks like problem solved!


jason-rice:
> would you include a test, please? it could be a unit test or integration test related to this.

PR #884, opened 2020-04-23
TPD-598 - added null check in EmailFinanceInformation method for missing from address

jrred:
In the Emailer.cs file, the method EmailFinanceInformation(MailAddress fromaddress, Person p, List<MailAddress> list, string subject, string body) would throw an exception if the fromaddress parameter was null (in ManagedGiving, it will be null if Util.TryGetMailAddress() doesn't find the email address).  I added a null check for that.

I also added a unit test for this method, SendEmailTests.cs > Send_Email_Without_From_Address_Does_Not_Throw_Exception()

### DIFF
--- a/UnitTests/CmsDataTests/Email/SendEmailTests.cs
+++ b/UnitTests/CmsDataTests/Email/SendEmailTests.cs
@@ -14,15 +14,17 @@ namespace CmsDataTests.Email
     [Collection(Collections.Database)]
     public class SendEmailTests : DatabaseTestBase
     {
-        [Fact]
-        public void Send_Email_Without_From_Address_Does_Not_Throw_Exception()
+        [Theory]
+        [InlineData(1, true)]
+        [InlineData(-1, true)]
+        public void Email_Finance_Information_Without_From_Address_No_Exception(int peopleId, bool expected)
         {
             bool success = true;
 
             try
             {
                 var from = Util.TryGetMailAddress("");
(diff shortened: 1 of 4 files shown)


jrred:
https://touchpointsoftware.atlassian.net/browse/TPD-598

UnitTests/CmsDataTests/Email/SendEmailTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UtilityExtensions;
+using Xunit;
+
+namespace CmsDataTests.Email
+{
+    [Collection(Collections.Database)]
+    public class SendEmailTests : DatabaseTestBase
```
brondavies, line 15: Naming convention of the test class (and file) should match the class being tested e.g. `CMSDataContextTests`  you might have to write tests in an existing class if it's in the same namespace.

UnitTests/CmsDataTests/Email/SendEmailTests.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        [Fact]
+        public void Send_Email_Without_From_Address_Does_Not_Throw_Exception()
```
brondavies, line 18: Naming convention for test methods should match (somewhat closely) the method being tested e.g. `EmailFinanceInformationTest`.  You might consider making this method use `[Theory]` and `[InlineData]` to test both cases (null and valid data)

PR #889, opened 2020-04-29
TPD-892 - Fixed Bundle creation redirect issue and added unit test

jrred:
Previously when creating a new Bundle (Administration > Bundles > click "Create New Bundle" button > fill out fields > click "Save"), the page would just redirect back to itself, instead of advancing to the confirmation page.  This was due to the querystring in the redirect somehow resetting the "create" parameter to "true", even if explicitly set to false in the redirect statement executed right after the Bundle is saved.  So I added a "sentinel" using TempData to maintain the state of "Bundle Created" to prevent the faulty redirect.
touchpointsoftware.atlassian.net/browse/TPD-892

### DIFF
--- a/CmsWeb/Areas/Finance/Controllers/BundleController.cs
+++ b/CmsWeb/Areas/Finance/Controllers/BundleController.cs
@@ -22,7 +22,12 @@ namespace CmsWeb.Areas.Finance.Controllers
         public ActionResult Index(int id, bool? create, bool? edit)
         {
             var m = new Models.BundleModel(id, CurrentDatabase);
-            ViewBag.createbundle = create;
+
+            if (TempData["BundleCreated"] == null)
+                ViewBag.createbundle = create;
+
+            TempData["BundleCreated"] = "True";
+
             ViewBag.editbundle = edit;
             if (User.IsInRole("FinanceDataEntry") && m.BundleStatusId != BundleStatusCode.OpenForDataEntry)
             {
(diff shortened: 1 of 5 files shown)


CmsWeb/Areas/Finance/Controllers/BundleController.cs
```diff
@@ -22,7 +22,12 @@ public BundleController(IRequestManager requestManager) : base(requestManager)
         public ActionResult Index(int id, bool? create, bool? edit)
         {
             var m = new Models.BundleModel(id, CurrentDatabase);
-            ViewBag.createbundle = create;
+
+            if (TempData["BundleCreated"] == null)
```
brondavies, line 26: @jrred  How does this fix the problem?  We don't reference TempData anywhere (anymore).

jrred:
@brondavies Fixed a couple of additional navigation problems. Link to edit an existing bundle and canceling out of edit mode were not working. Got those fixed.

CmsWeb/Web.config
```diff
@@ -45,7 +45,7 @@
     <!-- use (local) or .\SQLEXPRESS -->
     <add key="dbserver" value="" />
     <!--Second half of db name like CMS_bellevue-->
-    <add key="host" value="" />
+    <add key="host" value="myBvcms" />
```
brondavies, line 48: You'll have to undo this

CmsWeb/Web.config
```diff
@@ -1033,7 +1033,7 @@
   <system.net>
     <mailSettings>
       <smtp deliveryMethod="SpecifiedPickupDirectory">
-        <specifiedPickupDirectory pickupDirectoryLocation="c:\email" />
+        <specifiedPickupDirectory pickupDirectoryLocation="c:\dev\pursuant\email" />
```
brondavies, line 1036: this too

jrred:
web.config changes reverted.

PR #894, opened 2020-05-01
TPD-910 - added thumbnail creation to Example Template

jrred:
The Example Template (one to the far right) has the property IsUnLayer = true, unlike the other three templates.  This causes UnLayerCompose.cshtml to be rendered, rather than EditTemplate.cshtml like the others.

UnLayerCompose.cshtml calls DisplayController.cs > SaveUnlayerTemplate() when the “Save Template” button is clicked.  The other three templates call DisplayController.cs > ContentUpdate() instead.  That is why they all work (thumbnail image is created and displayed) whereas the Example Template link isn’t working.

So I added the image creation code to SaveUnlayerTemplate like the ContentUpdate method has.

Jira card: https://touchpointsoftware.atlassian.net/secure/RapidBoard.jspa?rapidView=6&projectKey=TPD&selectedIssue=TPD-892&assignee=5e9da7220d58350c2be09a8f

[touchpointsoftware.atlassian.net/browse/TPD-910](https://touchpointsoftware.atlassian.net/browse/TPD-910)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ CmsWeb/StagingConnectionStrings.config
 cassette-cache/
 CmsDataSqlClr.dbmdl
 CmsWeb/Upload/
+Web.config
 Web.*.config
 CmsDataSqlClr_.*.sqlproj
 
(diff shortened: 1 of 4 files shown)


jrred:
TPD-910

Added SaveUnlayerTemplate_Should_Create_Thumbnail unit test

.gitignore
```diff
@@ -64,6 +64,7 @@ CmsWeb/StagingConnectionStrings.config
 cassette-cache/
 CmsDataSqlClr.dbmdl
 CmsWeb/Upload/
+Web.config
```
brondavies, line 67: sorry, we can't ignore the web.config - there will be changes sometimes that have to be checked in.  Just make sure you don't check in changes specific to your environment 👍

PR #897, opened 2020-05-05
TPD-1003 Update blog widget to remove unicode characters

timdavila:
https://touchpointsoftware.atlassian.net/browse/TPD-1003


brondavies:
Is this the same code we updated everyone with already?

timdavila:
> Is this the same code we updated everyone with already?

Yes

PR #934, opened 2020-05-30
Fix duplicate cellphone problem incoming sms

davidcarroll:
And fix Notice Email


davidcarroll:
Good point @hkouns. 
I will take your advice and commit a change as soon as my internet is back up.

PR #955, opened 2020-06-23
Label when viewing does not match label when editing

jkrrv:
Also, doesn't really make sense.


brondavies:
@jkrrv  https://github.com/bvcms/bvcms/pull/973 replaces this PR and has the fix we agreed on internally for this issue.  Thanks for contributing!

PR #974, opened 2020-07-09
TPD-908 Fix exception issue in Pushpay Controller

dahnbaez:
https://touchpointsoftware.atlassian.net/secure/RapidBoard.jspa?rapidView=5&projectKey=TPD&selectedIssue=TPD-908&assignee=5e19301c85a8c90ecaca7d44
[touchpointsoftware.atlassian.net/browse/TPD-908](https://touchpointsoftware.atlassian.net/browse/TPD-908)


CmsWeb/Areas/Setup/Controllers/PushpayController.cs
```diff
@@ -308,7 +308,6 @@ public async Task<ActionResult> ProcessPayment(string paymentToken, string sr)
             }
             catch (Exception ex)
             {
-                ErrorSignal.FromCurrentContext().Raise(ex);
```
brondavies, line 311: This is supposed to record the exception in ELMAH

dahnbaez: I added it with that in mind but it was throwing an error in pushpay database. I'm not sure why... Elmah says:
"System.ArgumentNullException: Value cannot be null.
Parameter name: context
at Elmah.ErrorSignal.FromContext(HttpContext context)"

PR #1036, opened 2020-09-05
Make background checks (renewals, anyway) scriptable

jkrrv:
The goal was to move a bunch of background check submission logic from CmsWeb/.../VolunteeringController.cs to CmsData so as to abstract it from the viewmodel context, and make it available to the Python model, thereby allowing us to run automated background check renewals.


jkrrv:
@brondavies I found this difficult to test thoroughly, but it looks like my attempted background check submissions went through.  Let me know if there's some suggested method for more thoroughly testing these features.

brondavies (approved):
Thanks - nice addition 👍

PR #1041, opened 2020-09-10
TPD-1394, TPD-1273 Member Export Fix

josearmandohf:
Updating CurrOrgMembers2 function.
Only one migration file.
[touchpointsoftware.atlassian.net/browse/TPD-1394](https://touchpointsoftware.atlassian.net/browse/TPD-1394)
[touchpointsoftware.atlassian.net/browse/TPD-1273](https://touchpointsoftware.atlassian.net/browse/TPD-1273)


CmsWeb/Models/HomeModel.cs
```diff
@@ -600,7 +568,92 @@ orderby o.OrganizationName
...
+                           where p.FirstName.StartsWith(first) || p.NickName.StartsWith(first) || p.MiddleName.StartsWith(first) || p.AltName.Contains(first)                           
+                           select p
+                        : from p in query
+                          where p.IsBusiness != true
+                          where p.LastName.StartsWith(last) || p.MaidenName.StartsWith(last) || p.AltName.Contains(last)
+                          where p.FirstName.StartsWith(first) || p.NickName.StartsWith(first) || p.MiddleName.StartsWith(first) || p.AltName.StartsWith(first)
+                          select p;
+                    break;
+            }
+
```
dahnbaez, line 654: It seems that CodeFactor detects this switch structure as a complex method. Perhaps we can put these queries into methods to avoid this CodeFactor warning.

dahnbaez (changes requested):

CmsWeb/Models/HomeModel.cs
josearmandohf: Thanks Danny, actually the intention of this method was to encapsulate the queries happening before in FastSearch method directly. What makes CodeFactor take this as a complex method was the use of the conditional operator "?:". I've updated that to be simple if conditions and that seems to do the magic with codefactor.

dahnbaez (approved):

UnitTests/SharedTestFixtures/MockPeople.cs
```diff
@@ -60,8 +61,11 @@ public static void DeleteMockPerson(CMSDataContext db, Person person)
             var family = db.Families.FirstOrDefault(f => f.FamilyId == person.FamilyId);
             db.PurgePerson(person.PeopleId);                
             db.SubmitChanges();
-            db.Families.DeleteOnSubmit(family);
-            db.SubmitChanges();
+            if (family.IsNotNull())
```
brondavies, line 64: family should really only be deleted if there are no people in the family. Use PurgePerson first and then check for `family.People.Count`

josearmandohf: Thanks @brondavies working on it...

PR #1075, opened 2020-10-02
TPD-1372, TPD-1142 ReCheck SMS Sent Status

dahnbaez:
https://touchpointsoftware.atlassian.net/secure/RapidBoard.jspa?rapidView=5&projectKey=TPD&selectedIssue=TPD-1372&assignee=5e19301c85a8c90ecaca7d44
[touchpointsoftware.atlassian.net/browse/TPD-1372](https://touchpointsoftware.atlassian.net/browse/TPD-1372)


brondavies (approved):

josearmandohf (changes requested):
Can you add unit tests?

UnitTests/CmsDataTests/Classes/Twilio/TwilioHelperTests.cs
```diff
@@ -0,0 +1,24 @@
...
+    [Collection(Collections.Database)]
+    public class TwilioHelperTests : DatabaseTestBase
+    {
+        [Fact]
+        public void Should_Not_Send_SMS_When_Empty_Credentials()
+        {
+            db.SetSetting("TwilioSID", "");
+            db.SetSetting("TwilioToken", "");
+
+            TwilioHelper.ProcessQueue(db, 1);
```
josearmandohf, line 21: This test does not assert anything. 

The ProcessQueue() method uses an iListID parameter that then help to get a list of SMSLists DB objects. If Process is successful then at the end in TwillioHelper class Line 314 smsList.Sent is set to true.

You should be asserting smsList.Sent = true. 

An Idea is to fetch or create a list of SMSLists and the SelectFirstOrDefault(), then pass that as iListID to execute and then assert. Something like the following...

var smslist = db.SMSLists.FirstOrDefault();
TwilioHelper.ProcessQueue(db, smslist.ID)
smsList.Sent.ShouldBe(False);

josearmandohf (changes requested):

josearmandohf (approved):
Approved

PR #1096, opened 2020-10-13
TPD-1415 Prevent Duplicate Transactions

davidcarroll:
Use Post, Redirect, Get (PRG) approach to avoid resubmitting a payment transaction.
[touchpointsoftware.atlassian.net/browse/TPD-1415](https://touchpointsoftware.atlassian.net/browse/TPD-1415)


CmsWeb/Code/CmsController.cs
```diff
@@ -74,6 +75,17 @@ public ViewResult PageMessage(string text, string title = "Error", string alert
...
+        {
+            RequestManager.SessionProvider.Add(name, value);
+        }
+        public string GetTempData(string name)
+        {
+            var ret = RequestManager.SessionProvider.Get<string>(name);
+            RequestManager.SessionProvider.Delete(name);
+            return ret;
+        }
+
```
brondavies, line 88: We have a pattern for this in [Util.Session.cs](https://github.com/bvcms/bvcms/blob/develop/UtilityExtensions/Extensions/Util.Session.cs#L258) - use that method so we can keep track of which keys are used for the session.

CmsWeb/Areas/OnlineReg/Controllers/Payment.cs
```diff
@@ -102,43 +101,59 @@ public ActionResult ProcessPayment(PaymentForm pf)
...
 
-            switch (ret.Route)
+            switch (routeModel.Route)
             {
                 case RouteType.ModelAction:
-                    ret.Model.SaveTicketingOrderId();
-                    return View(ret.View, ret.Model);
+                    routeModel.Model.SaveTicketingOrderId();
+                    AddTempData("RouteModel", routeModel.WriteXml());
+                    return Redirect("/OnlineReg/ProcessPayment");
```
brondavies, line 123: Is there a possibility that the registration could end up here if a payment is not needed on the registration?

CmsWeb/Areas/OnlineReg/Controllers/Payment.cs
```diff
@@ -102,43 +101,59 @@ public ActionResult ProcessPayment(PaymentForm pf)
...
                 case RouteType.AmtDue:
-                    ViewBag.amtdue = ret.AmtDue;
-                    return View(ret.View, ret.Transaction);
+                    AddTempData("RouteModel", routeModel.WriteXml());
+                    return Redirect("/OnlineReg/ProcessPayment");
```
brondavies, line 126: And here - is there a possibility that the registration could end up here if a payment is not needed on the registration?

brondavies, line 126: `ProcessExternalPayment` won't necessarily redirect to this URL

PR #1125, opened 2020-10-23
TPD-1580, TPD-1549 Fix Pledge Tab, Giving tab

josearmandohf:
Issues:

My Pledges toggle does not work as expected

Pledged column is not needed, shift other date in row 2+ to the right one

Total projected calculation is not being done (Check w/Tim)

When recurring pledges is turned on hide "Pledge Detail" for all non-finance (any type) users

Make the new “Make a Pledge” button hidden behind a setting >> finance >> pledges >> ShowMakeAPledgeButton (true/false), default should be false.

[touchpointsoftware.atlassian.net/browse/TPD-1580](https://touchpointsoftware.atlassian.net/browse/TPD-1580)
[touchpointsoftware.atlassian.net/browse/TPD-1549](https://touchpointsoftware.atlassian.net/browse/TPD-1549)


timdavila (approved):
Looks good 👍

PR #1127, opened 2020-10-27
Python Model Adjustments

jkrrv:
Adding a function for non-JSON POSTing.  (The primary impetus is that apparently the JSON serializer of RestPostJson doesn't handle recursion very well/at all.  This function allows an alternate serializer, or non-JSON data.)

Also, making PR #1036 do what it was meant to do all along.


jkrrv:
@brondavies

jkrrv:
@brondavies ???

brondavies (approved):

PR #1142, opened 2020-11-02
make search for fund without fall back accessible in the python api

timdavila:
to support https://touchpointsoftware.atlassian.net/browse/TPD-1585


josearmandohf (approved):
Approved

PR #1146, opened 2020-11-03
TPD-1568 Donor Covers Fee Transactions Log

jason-rice:
Finished!
[touchpointsoftware.atlassian.net/browse/TPD-1568](https://touchpointsoftware.atlassian.net/browse/TPD-1568)


dahnbaez (changes requested):
There is a message from CodeFactor that needs to be addressed. I think you only need to remove the line commented at TransactionsModel.cs#L191.

dahnbaez (approved):

PR #1156, opened 2020-11-06
TPD-1611 Delete Giving Pages

jason-rice:
done
[touchpointsoftware.atlassian.net/browse/TPD-1611](https://touchpointsoftware.atlassian.net/browse/TPD-1611)


CmsWeb/src/components/Giving/GivingManagement/GivingPageIndex.vue
```diff
@@ -1,162 +1,275 @@
...
+}
+.swal-title {
+  font-size: 12px;
+}
+.swal-text {
+  font-size: 12px;
+}
+.delete-button {
+  font-size: 14px;
+}
```
brondavies, line 274: This is something that should apply the same style everywhere and not just in this component

CmsWeb/Areas/Giving/Controllers/GivingManagementController.cs
```diff
@@ -79,13 +80,12 @@ public ActionResult Create(GivingPageViewModel viewModel)
             return Json(result, JsonRequestBehavior.AllowGet);
         }
 
-        [HttpPost]
```
brondavies, line 82: We should probably keep the `HttpPost` designation if that's how this is accessed

brondavies (changes requested):

CmsWeb/src/components/Giving/GivingManagement/GivingPageIndex.vue
jason-rice: deleted this

CmsWeb/Areas/Giving/Controllers/GivingManagementController.cs
jason-rice: put this attribute back

josearmandohf (approved):
Changes required were addressed.

PR #1157, opened 2020-11-09
TPD-1653 Limit People Dropdown

jason-rice:
finished
[touchpointsoftware.atlassian.net/browse/TPD-1653](https://touchpointsoftware.atlassian.net/browse/TPD-1653)


josearmandohf (approved):
It seems legit 👍

PR #1163, opened 2020-11-12
TPD-1614 Adding db function SplitIntsKey to order by the Key instead of peopleid.

josearmandohf:
[touchpointsoftware.atlassian.net/browse/TPD-1614](https://touchpointsoftware.atlassian.net/browse/TPD-1614)


CmsData/CmsData.csproj
```diff
@@ -1247,6 +1248,7 @@
     <Resource Include="Classes\OnlineRegSummaryText\Details2.xml">
       <SubType>Designer</SubType>
     </Resource>
+    <Content Include="Migrations\20201112_create_SplitIntsKey_function.sql" />
```
brondavies, line 1251: Reminder: Don't include migrations in the project file

CmsWeb/Areas/Giving/Models/GivingPage/GivingPageModel.cs
```diff
@@ -71,8 +71,9 @@ public List<GivingPageItem> GetGivingPages(int id = 0)
                                               }).SingleOrDefault(),
                                 TopText = gp.TopText,
                                 ThankYouText = gp.ThankYouText,
-                                OnlineNotifyPerson = (from np in db.People
-                                                      where onp.Contains(np.PeopleId)
+                                OnlineNotifyPerson = (from p in onp
+                                                      join np in db.People on p.ValueX equals np.PeopleId
+                                                      orderby p.TokenID
```
brondavies, line 76: This should fix the problem of ordering without the new function that is essentially the same as SplitInts

brondavies (changes requested):
I don't think we need the new SplitIntsKey function

CmsData/CmsData.csproj
josearmandohf: Right, removing it...

brondavies (approved):

PR #1165, opened 2020-11-13
TPD-1478 Add Person in Regs Fix

dahnbaez:
https://touchpointsoftware.atlassian.net/browse/TPD-1478?atlOrigin=eyJpIjoiZTExZDQ2NTQ5NDEyNDEzOTg0Zjc0N2Y5MjNhOWY2NDciLCJwIjoiaiJ9


CmsData/Organization/Organization.cs
```diff
@@ -457,6 +457,7 @@ public OrganizationExtra GetExtraValue(string field, bool create = true)
                 {
                     OrganizationId = OrganizationId,
                     Field = field,
+                    BitValue = true
```
brondavies, line 460: The only problem with this is that if you are calling GetExtraValue for an EV that should be a different data type, you'll have multiple values on the extra value which can cause problems. I think it would be better to change the behavior here https://github.com/bvcms/bvcms/blob/develop/CmsWeb/Areas/OnlineReg/Models/OnlineReg/HelperMethods.cs#L250

brondavies (changes requested):

brondavies (approved):
Just need to fix the test

PR #1182, opened 2020-11-27
TPD-1728 Fix Duplicate Donation OnlineReg

davidcarroll:
See commented out code on line 183 in Confirm.cs
[touchpointsoftware.atlassian.net/browse/TPD-1728](https://touchpointsoftware.atlassian.net/browse/TPD-1728)


josearmandohf (approved):
I've tested this with all cases I have for registration and it seems to be correct.

PR #1191, opened 2020-12-03
Add Validation to scratchpad on Birthday

kirkentblk:
Example demonstrating ability to analyze and code in solution...do not merge (pulled on a thread and got a sweater).

To replicate, open UI-> Login -> People -> New Search -> Select Birthday from the list -> type Fred in the text box.
After fix only MM, MM/DD, YYYY, or MM/DD/YYYY will be accepted.

Added validation to scratchpad custom searches when Birthday is selected to be in one of the listed formats.  Updated UI to display error message on page.  Added unit tests on QueryModel changes.


CmsWeb/ConnectionStrings.config
```diff
@@ -1,4 +1,4 @@
 <connectionStrings>
-  <add name="CMS" connectionString="Data Source=(local);Integrated Security=True" />
-  <add name="Elmah.Sql" connectionString="Data Source=(local);Initial Catalog=ELMAH;Integrated Security=True" />
```
brondavies, line 3: This change should be reverted

CmsWeb/Web.config
```diff
@@ -46,7 +46,7 @@
     <!-- use (local) or .\SQLEXPRESS -->
     <add key="dbserver" value="" />
     <!--Second half of db name like CMS_bellevue-->
-    <add key="host" value="" />
```
brondavies, line 49: This should be reverted too

CmsWeb/package.json
```diff
@@ -72,7 +72,6 @@
         "last 2 versions",
         "not dead"
     ],
-    "_id": "cmsweb@1.0.0",
```
brondavies, line 75: This should be reverted too

UnitTests/CMSWebTests/Areas/Search/Models/Query/QueryModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using CmsWeb.Areas.Search.Models;
+using UtilityExtensions;
+using System.Linq;
+using System;
+using System.Collections.Generic;
+
+namespace CMSWebTests.Areas.Search.Models.Query
+{
+    [Collection(Collections.Database)]
+    public class QueryModelTests : IDisposable
```
brondavies, line 16: In this case, this test class can be greatly simplified by making `IsValidBirthdayQuery` a `static` method, then you don't need to instantiate the object to test it and can do away with the IDisposable implementation and other non-essentials (db, model, user, etc).

kirkentblk: I'll update tonight and submit by morning.

